PR #33, opened 2021-01-29
Automatically correct n in simpsons boole

htoftevaag:
Updated Simpson's and Boole's so that it gives a warning if the N per dim is wrong, and then the N per dim is automatically corrected. 

Added description to torchquad/integration/utils.py.
Fixed a small mistake (N=2 but said it should be 1000) in Trapezoid.

And lastly, changed typos (mostly punctuation), including in simpson_test and boole_test and Trapezoid.

### DIFF
--- a/torchquad/integration/boole.py
+++ b/torchquad/integration/boole.py
@@ -17,33 +17,28 @@ class Boole(BaseIntegrator):
     def __init__(self):
         super().__init__()
 
-    def integrate(self, fn, dim, N=5, integration_domain=None):
-        """Integrates the passed function on the passed domain using Boole's rule
+    def integrate(self, fn, dim, N=None, integration_domain=None):
+        """Integrates the passed function on the passed domain using Boole's rule.
 
         Args:
             fn (func): The function to integrate over.
             dim (int): Dimensionality of the integration domain.
-            N (int, optional): Number of sample points to use for the integration. N has to be such as N^(1/dim) - 1 % 4 == 0. Defaults to 5 for dim = 1.
-            integration_domain (list, optional): Integration domain. Defaults to [-1,1]^dim.
+            N (int, optional): Total number of sample points to use for the integration. N has to be such that N^(1/dim) - 1 % 4 == 0. Defaults to 5 points per dimension if None is given.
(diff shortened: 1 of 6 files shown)


torchquad/integration/boole.py
```diff
@@ -77,6 +78,37 @@ def integrate(self, fn, dim, N=5, integration_domain=None):
...
+        """Adjusts the total number of points to a valid number, i.e. N satisfies N^(1/dim) - 1 % 4 == 0. 
+
+        Args:
+            dim (int): Dimensionality of the integration domain.
+            N (int): Total number of sample points to use for the integration.
+            
+        Returns:
+            int: An N satisfying N^(1/dim) - 1 % 4 == 0.
+        """
+        Nperdim = int(N ** (1.0 / dim) + 1e-8)
```
gomezzz, line 95: please rename to n_per_dim (to match with our naming conventions)

torchquad/integration/simpson.py
```diff
@@ -80,7 +76,37 @@ def integrate(self, fn, dim, N=3, integration_domain=None):
...
+        """Adjusts the current N to an odd integer >=3, if N is not that already. 
+
+        Args:
+            dim (int): Dimensionality of the integration domain.
+            N (int): Total number of sample points to use for the integration.
+            
+        Returns:
+            int: An odd N >3.
+        """
+        Nperdim = int(N ** (1.0 / dim) + 1e-8)
```
gomezzz, line 93: as before

torchquad/tests/simpson_test.py
```diff
@@ -32,7 +32,7 @@ def test_integrate():
     # 3D Tests
     N = 1030301  # N = 101 per dim
```
gomezzz, line 33: I think in the other test you changed it so that it breaks if the N-finding method doesn't work correctly, right? Maybe do that here as well?

gomezzz (changes requested):

torchquad/tests/simpson_test.py
htoftevaag: Copy that

gomezzz (approved):

PR #34, opened 2021-01-29
Added _runtime to integration_grid

GabrieleMeoni:
Separating runtime measurements for grid creation, function evaluation, and total computing.
(WIP: the notebook is memory hungry.)


notebooks/Benchmark_notebook.ipynb
```diff
@@ -54,20 +54,33 @@
    "outputs": [],
    "source": [
     "e = 2.7182818284590452353602874 #hardcoded to avoid dependence on torch / numpy\n",
+    "runtime_f = 0 #Runtime global variable\n",
```
gomezzz, line 57: I think you need to set this 0 again in the `evaluate` function? Otherwise it will never be reset, if I am not mistaken

notebooks/Benchmark_notebook.ipynb
```diff
@@ -54,20 +54,33 @@
...
     "def test_function_1d(x):\n",
     "    \"\"\" As a test function in 1D we use e^x\n",
     "    \"\"\"\n",
-    "    return e**x\n",
+    "    global runtime_f\n",
```
gomezzz, line 61: I don't get the point of using a global variable if you overwrite it every time? Wouldn't it be better to accumulate all function runtimes with

```python
start_f = perf_counter()
...
global runtime_f
runtime_f = runtime_f + (perf_counter() - start)
```

and then take that value at the end?

gomezzz: Further, maybe you can create a function

```python
def track_fn_runtime(start):
global runtime_f
runtime_f = runtime_f + (perf_counter() - start)
```

which improves readability and saves you a few lines?

notebooks/Benchmark_notebook.ipynb
```diff
@@ -97,19 +110,23 @@
...
     "    \"\"\"Wrapper for scipy trapezoid method, creates grid as well\n",
     "    \"\"\"\n",
-    "    eval_points = torchquad.IntegrationGrid(N,domain).points.cpu().numpy().transpose()\n",
-    "    y = fn(eval_points)\n",
-    "    return trapz(y,eval_points)[0]\n",
+    "    grid = torchquad.IntegrationGrid(N,domain)\n",
+    "    eval_points = grid.points.cpu().numpy().transpose()\n",
+    "    torch.cuda.synchronize()\n",
+    "    y = test_fn(eval_points)\n",
+    "    return trapz(y,eval_points)[0], grid._runtime, runtime_f\n",
```
gomezzz, line 120: runtime_fn is a global variable, you don't need to return it

notebooks/Benchmark_notebook.ipynb
```diff
@@ -147,35 +164,38 @@
...
     "    \"\"\"3D trapezoid using quadpy, note that it will perform 8*N evals\n",
     "    \"\"\"\n",
+    "    start_grid = perf_counter()\n",
     "    cubes = _quadpy_helper(N,domain)\n",
     "#     plot_cubes(cubes)\n",
+    "    runtime_grid = perf_counter() - start_grid\n",
     "    scheme = quadpy.c3.product(quadpy.c1.newton_cotes_closed(1))\n",
-    "    val = scheme.integrate(fn,cubes)\n",
-    "    return 0.5*sum(val) #this 0.5 is very odd because it implies that we integrate each domain twice, however we don't?\n",
+    "    return 0.5 * sum(scheme.integrate(test_fn,cubes)), runtime_grid, runtime_f #this 0.5 is very odd because it implies that we integrate each domain twice, however we don't?\n",
```
gomezzz, line 175: why was this merged into one line? it is now too long to be read

notebooks/Benchmark_notebook.ipynb
```diff
@@ -147,35 +164,38 @@
...
+    "    start_grid = perf_counter()\n",
     "    cubes = _quadpy_helper(N,domain)\n",
     "#     plot_cubes(cubes)\n",
     "    scheme = quadpy.c3.product(quadpy.c1.newton_cotes_closed(2))\n",
-    "    val = scheme.integrate(fn,cubes)\n",
-    "    return 0.5*sum(val) #this 0.5 is very odd because it implies that we integrate each domain twice, however we don't?\n",
-    "    \n",
-    "def torchquad_trapezoid(fn,dim,N,domain):\n",
+    "    runtime_grid = perf_counter() - start_grid\n",
+    "    return 0.5*sum(scheme.integrate(test_fn,cubes)), runtime_grid, runtime_f #this 0.5 is very odd because it implies that we integrate each domain twice, however we don't?\n",
```
gomezzz, line 186: as above

notebooks/Benchmark_notebook.ipynb
```diff
@@ -147,35 +164,38 @@
...
+    "\n",
+    "def torchquad_trapezoid(test_fn,dim,N,domain):\n",
     "    \"\"\"Wrapper for scipy simpsons method, creates grid as well\n",
     "    \"\"\"\n",
     "    tp = torchquad.Trapezoid()\n",
-    "    return tp.integrate(fn,dim,N,domain).item()\n",
+    "    return tp.integrate(test_fn,dim,N,domain).item(), tp._grid._runtime, runtime_f\n",
```
gomezzz, line 192: runtime_f is global

notebooks/Benchmark_notebook.ipynb
```diff
@@ -147,35 +164,38 @@
...
     "    \n",
-    "def torchquad_simpsons(fn,dim,N,domain):\n",
+    "def torchquad_simpsons(test_fn,dim,N,domain):\n",
     "    \"\"\"Wrapper for scipy simpsons method, creates grid as well\n",
     "    \"\"\"\n",
     "    sp = torchquad.Simpson()\n",
-    "    return sp.integrate(fn,dim,N,domain).item()"
+    "    return sp.integrate(test_fn,dim,N,domain).item(), sp._grid._runtime, runtime_f"
```
gomezzz, line 198: runtime_f is global

notebooks/Benchmark_notebook.ipynb
```diff
@@ -197,23 +217,26 @@
...
-    "    return result,time\n",
+    "    return result, grid_runtime, fn_runtime, time -(grid_runtime + fn_runtime)\n",
     "\n",
     "def benchmark(integrator,fn,dim,N,domain, iterations=10):\n",
     "    \"\"\" Evaluates given Integrator once as warmup, then #iterations times\n",
     "    \"\"\"\n",
-    "    results,times = [],[]\n",
+    "    results = []\n",
+    "    grid_runtime, fn_runtime, integration_time = 0, 0, 0\n",
     "    #Do three calls as warmup (to allow CPU / GPU to use caching and increase frequency)\n",
```
gomezzz, line 229: not necessarily "three" , maybe change to "some"?

notebooks/Benchmark_notebook.ipynb
```diff
@@ -197,23 +217,26 @@
     "    \"\"\"Evaluates given integrator and returns result and runtime\n",
     "    \"\"\"\n",
     "    start = perf_counter()\n",
-    "    result = integrator(fn,dim,N,domain)\n",
+    "    result, grid_runtime, fn_runtime = integrator(fn,dim,N,domain)\n",
     "    time = perf_counter() - start\n",
-    "    return result,time\n",
+    "    return result, grid_runtime, fn_runtime, time -(grid_runtime + fn_runtime)\n",
```
gomezzz, line 222: This is a good place to actually return runtime_f. It may be good though to make a deepcopy of it to avoid accidental modification (global variable and all...)

notebooks/Benchmark_notebook.ipynb
```diff
@@ -197,23 +217,26 @@
...
     "    \"\"\" Evaluates given Integrator once as warmup, then #iterations times\n",
     "    \"\"\"\n",
-    "    results,times = [],[]\n",
+    "    results = []\n",
+    "    grid_runtime, fn_runtime, integration_time = 0, 0, 0\n",
     "    #Do three calls as warmup (to allow CPU / GPU to use caching and increase frequency)\n",
     "    for i in range(WARMUP_CALLS): \n",
     "        evaluate(integrator,fn,dim,N,domain)\n",
     "    #Do iteration benchmark calls\n",
     "    for it in range(iterations):\n",
```
gomezzz, line 233: Okay, so here, I think, you need to set `runtime_f = 0` (warmup calls should not be part of the measured time of course)

gomezzz: (before the loop I mean)

notebooks/Benchmark_notebook.ipynb
```diff
@@ -274,13 +301,52 @@
     "torchquad.plot_convergence(steps_per_method,values,ground_truth_1d,names)"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Plotting the total runtime for integration (creation of integration grid + function evaluation + integration computing)."
```
gomezzz, line 308: I don't think that is true. Because above in `evaluate` you subtracted grid and runtime from integration time (which doesn't actually work at the moment , because you don't measure total evaluation time of function calls atm but only one function  call)

notebooks/Benchmark_notebook.ipynb
```diff
@@ -274,13 +301,52 @@
...
+    "Plotting the total runtime for function evaluation."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "torchquad.plot_runtime(steps_per_method,fn_runtime,names, y_axis_name=\"fn eval - runtime [s]\")"
```
gomezzz, line 349: Please a bit more descriptive: fn eval - runtime [s] -> Function evaluations

notebooks/Benchmark_notebook.ipynb
```diff
@@ -335,13 +407,52 @@
     "torchquad.plot_convergence(steps_per_method,values,ground_truth_3d,names)"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Plotting the total runtime for integration (creation of integration grid + function evaluation + integration computing)."
```
gomezzz, line 414: as above

notebooks/Benchmark_notebook.ipynb
```diff
@@ -335,13 +407,52 @@
...
+    "Plotting the total runtime for function evaluation."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "torchquad.plot_runtime(steps_per_method,fn_runtime,names,y_axis_name=\"fn eval - runtime [s]\")"
```
gomezzz, line 455: as above

torchquad/integration/integration_grid.py
```diff
@@ -3,6 +3,7 @@
 
 logger = logging.getLogger(__name__)
 
+from time import perf_counter
```
gomezzz, line 6: pls move this above the logger declaration

gomezzz (changes requested):

notebooks/Benchmark_notebook.ipynb
gomezzz: I think this was not changed?

notebooks/Benchmark_notebook.ipynb
```diff
@@ -302,19 +401,23 @@
...
-    "steps = np.asarray([3,5,7,9,11,15])**3\n",
     "for method in methods:\n",
-    "    cur_values,cur_time = [],[]\n",
+    "    # Numbers of evals / integration points to test\n",
+    "    steps = np.asarray([3,5,7,9,11,15])**3\n",
+    "    cur_values, cur_grid_runtime, cur_fn_runtime, cur_integration_time = [],[],[],[]\n",
     "    for step in tqdm(steps):\n",
     "        if method in [quadpy_trapezoid_3d,quadpy_simpson_3d]:\n",
-    "            vals,time_measuresments = benchmark(method,test_function_3d_quadpy,3,int(step),test_domain_3d)\n",
+    "            vals,grid_runtime_iter, fn_runtime_iter, integration_time_iter = benchmark(method,test_function_3d_quadpy,3,int(step),test_domain_3d)\n",
```
gomezzz, line 413: Gives me 
image

gomezzz (changes requested):

notebooks/Benchmark_notebook.ipynb
gomezzz:
```python
val = scheme.integrate(fn,cubes) #this 0.5 is very odd because it implies that we integrate each domain twice, however we don't?
 return val, runtime_grid, runtime_f 
```

gomezzz (approved):

PR #40, opened 2021-02-19
Added logo to the left bar.

GabrieleMeoni:
Removed logo from index page.


htoftevaag:
Looks good, but we should fix the blue banner at a later time.

PR #41, opened 2021-02-19
Changed background of logo to gray-beige

htoftevaag:
Just a shading difference, to make the logo a bit prettier (without visible edges)


GabrieleMeoni (commented):
The new image is fine, but I struggle to see a difference between the new and the old image.

htoftevaag:
It has a slight shade of beige-grey, which blends in with the other background there

GabrieleMeoni (approved):

PR #52, opened 2021-04-09
Adding manual JOSS pdf compilation possibility

htoftevaag:
Now we can compile the paper manually, also for branches


GabrieleMeoni (commented):
I am merging to test the action with the updated draft-pdf.yml file.

PR #62, opened 2021-05-07
Support float and double precision switching

htoftevaag:
Notes:

- Right now, if the user writes _anything_ other than "double", it will be `float` by default
- The user is asked to enable CUDA before setting precision 
- If the user, however, sets precision before enabling CUDA, there is a redundancy in the enable_cuda code where it sets the precision to `cuda.Float32` automatically.


htoftevaag:
Question: If neither enable_cuda nor set_precision called, `float` will automatically be used?

gomezzz:
> 
> 
> Question: If neither enable_cuda nor set_precision called, `float` will automatically be used?

from https://pytorch.org/docs/stable/generated/torch.set_default_tensor_type.html

> The default floating point tensor type is initially torch.FloatTensor.

So float, but on the CPU. (not torch.cuda.FloatTensor )

torchquad/utils/set_precision.py
```diff
@@ -0,0 +1,30 @@
+import torch
+import os
+
+import logging
+
+logger = logging.getLogger(__name__)
```
gomezzz, line 6: You import the logger but don't log anything -> either remove or add logging

torchquad/utils/set_precision.py
```diff
@@ -0,0 +1,30 @@
...
+def set_precision(data_type="float"):
+    """This function allows the user to set the default precision. Call before declaring any variables!
+    NB: Remember to initialize CUDA first if GPU compatibility is desired.
+
+    Args:
+        data_type (string, optional): Data type to use, either "float" or "double". Defaults to "float".
+    """
+
+    if torch.cuda.is_initialized():
+        if data_type.lower() == "double":
```
gomezzz, line 18: I think I would differentiate between "float", "double" and if it is neither throw an error. Currently if I pass "complex" which makes no sense, it will just silently go to float

torchquad/utils/set_precision.py
```diff
@@ -0,0 +1,30 @@
...
+            print("CUDA is initialized. Setting default tensor type to cuda.Float64")
```
gomezzz, line 19: Probably we should start using logging instead of print, I think.  (Then you can also keep the logger!) (It is buggy in notebooks at the moment though, have created a ticket for that)

gomezzz: I'd log this with logger.warn("") probably just to make sure it is displayed by default? (If people want high-accuracy integrals accidentally switching everything to float may be very inconvenient)

gomezzz (changes requested):
See comments. Functionality is there, just wanna streamline it more :)

torchquad/utils/enable_cuda.py
```diff
@@ -18,6 +18,7 @@ def enable_cuda(device=0):
         logger.info("__CUDNN VERSION:" + str(torch.backends.cudnn.version()))
         logger.info("__Number of CUDA Devices:" + str(torch.cuda.device_count()))
         logger.info("Active CUDA Device: GPU" + str(torch.cuda.current_device()))
+        # Is setting the default tensor type to cuda.Float32 here a good redundancy? I think so
```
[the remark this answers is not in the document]
htoftevaag: I thought about the same idea, but is it annoying for the user to have to set the precision when they want to enable CUDA?

torchquad/utils/set_precision.py
```diff
@@ -0,0 +1,49 @@
...
+            )
+            torch.set_default_tensor_type(torch.cuda.DoubleTensor)
+        elif data_type.lower() == "float":
+            logger.warn(
+                "Setting default tensor type to cuda.Float32 (CUDA is initialized)."
+            )
+            torch.set_default_tensor_type(torch.cuda.FloatTensor)
+        else:
+            logging.error(
+                f"{data_type} is not a valid data type. Only float or double are supported.
```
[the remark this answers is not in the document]
htoftevaag: Copy that. Friday afternoon sloppy work from my side

torchquad/utils/enable_cuda.py
```diff
@@ -2,6 +2,7 @@
 import os
 
 import logging
+from utils.set_precision import set_precision
```
gomezzz: Fails with 
grafik

torchquad/utils/set_precision.py
```diff
@@ -26,8 +26,9 @@ def set_precision(data_type="float"):
             torch.set_default_tensor_type(torch.cuda.FloatTensor)
         else:
             logging.error(
-                f"{data_type} is not a valid data type. Only float or double are supported.
-                Setting default tensor type to cuda.Float32 (CUDA is initialized)."
+                data_type,
+                "is not a valid data type. Only float or double are supported. "
```
gomezzz: Like this there is a white space missing. Will look like this:

`floatis not a valid data type. Only float or double are supported. `

torchquad/utils/enable_cuda.py
gomezzz: when I try to do
`import torchquad`

gomezzz:
@htoftevaag I took the liberty to quickly fix outstanding cmments myself.

gomezzz (approved):

PR #66, opened 2021-05-07
Add specific version numbers for all requirements in environment.yml

gomezzz:


htoftevaag (approved):
All version numbers checked - looks good.

PR #63, opened 2021-05-07
Add cuda support to unit tests

gomezzz:
This will require a few more PRs to be merged before it (#59 #62 and one on VEGAS vectorization.) Don't merge / review yet!

PR #67, opened 2021-05-07
Vectorize vegas

gomezzz:
First merge #62 (due to a cross dependency this PR very slightly overlaps with #63).

### DIFF
--- a/torchquad/integration/vegas.py
+++ b/torchquad/integration/vegas.py
@@ -82,7 +82,7 @@ class VEGAS(BaseIntegrator):
         # Initialize the adaptive VEGAS map,
         # Note that a larger number of intervals may lead to problems if only few evals are allowed
         # Paper section II B
-        self.map = VEGASMap(self._dim, self._integration_domain, N_intervals=100)
+        self.map = VEGASMap(self._dim, self._integration_domain, N_intervals=5)
 
         # Initialize VEGAS' stratification
         # Paper section III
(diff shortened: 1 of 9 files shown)


torchquad/integration/vegas.py
```diff
@@ -134,7 +140,7 @@ def integrate(
                 if chi2 / 5.0 < 1.0:
                     self._starting_N = torch.minimum(
                         torch.tensor(self._starting_N + self._N_increment),
-                        self._starting_N * torch.sqrt(acc / eps_rel),
+                        self._starting_N * torch.sqrt(acc / (eps_rel + 1e-8)),
```
GabrieleMeoni, line 143: Isn't 1e-8 a bit high for an eps value?

GabrieleMeoni (changes requested):

torchquad/integration/vegas.py
```diff
@@ -82,7 +82,10 @@ def integrate(
         # Initialize the adaptive VEGAS map,
         # Note that a larger number of intervals may lead to problems if only few evals are allowed
         # Paper section II B
-        self.map = VEGASMap(self._dim, self._integration_domain, N_intervals=100)
+        N_intervals = max(2, self._N_increment // 10)  # for small N intervals set 2
```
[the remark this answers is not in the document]
gomezzz: See [https://www.w3schools.com/python/python_operators.asp](here). `//` is the floor operator.

torchquad/integration/vegas.py
gomezzz: It is actually rather small.

See row binary32 machine epsilon [https://en.wikipedia.org/wiki/Machine_epsilon#Values_for_standard_hardware_floating_point_arithmetics](here).

GabrieleMeoni (approved):

PR #72, opened 2021-05-28
Getting Release branch up to date for v0.2.0

gomezzz:
# Changelog
## Major
- Huge improvements to documentation
- Added VEGAS Enhanced
## Minor
- Added Boole's rule
- Migrated examples to docs
- Added JOSS Paper drafts
- Updated requirements
- Added option to set precision (float or double)
- Improved comments
- Added 10D unit test


torchquad/__init__.py
```diff
@@ -7,24 +7,30 @@
...
 
 from .utils.enable_cuda import enable_cuda
+from .utils.set_precision import set_precision
 
 __all__ = [
     "IntegrationGrid",
     "MonteCarlo",
     "Trapezoid",
     "Simpson",
+    "Boole",
```
GabrieleMeoni, line 28: Maybe we should report as major that we added boole? I don't see it in the changelog.

GabrieleMeoni (changes requested):
The Benchmark Notebook has not been removed in the end. It is on purpose?

torchquad/__init__.py
gomezzz: Done.

gomezzz:
> The Benchmark Notebook has not been removed in the end. It is on purpose?

We didn't have a ticket in it, but it's not in the pip / conda version any ways. Hence, I would remove it in 0.3?

GabrieleMeoni (approved):

PR #88, opened 2021-06-29
Code cleanup and fixed some issues

albernsrya:


gomezzz:
@albernsrya Are you still on this? :) 

If so, if we could work on getting the tests to pass. I'll also close the other individual PRs to clean up a little (I think the changes are all here, right?)

Also, note that this PR should merge into esa:develop not esa:main.

albernsrya:
hello @gomezzz  I've finished, thanks

PR #90, opened 2021-07-07
Documentation fixes, small typos

### LINKED ISSUE #89: Initial feedback from JOSS review
> 1. There's a grammatical mistake here: https://github.com/esa/torchquad/blob/2824f1ddacdccadf3a880851f7ddf1bc1f878a36/torchquad/integration/base_integrator.py#L73external should be either "needs to match" or "needs to be match to". In either case, since integration_domain is referred by variable name, ndim would be better to be explictly mention as well

>   2. The usage example is clear and it works, but it should include the enable_cuda() call explicitly so that the user knows, just reading the readme, how to use the GPU-capabilities of the library (which is one of the main points)

>   3. From the readme and rtd, it seems to work only with CUDA/nvidia. Is that the case? It should be mentioned explicitly in the readme and docs. Note that pytorch offers also a rocm version so users might expect it to work with their AMD cards.

We have received initial feedback from [JOSS reviewers ](https://github.com/openjournals/joss-reviews/issues/3439) which we should address.

gomezzz:
- Fixed some references to latest on rtd instead of main
- Added mention of no AMD support atm
- Added enable_cuda() to minimal example
- Added more comments to minimal example
- Small bugfix in minimal example
- Fixed typo in base_integrator

Closes #89

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 *** Based on https://github.com/othneildrew/Best-README-Template
 -->
 
-[![Documentation Status](https://readthedocs.org/projects/torchquad/badge/?version=latest)](https://torchquad.readthedocs.io/en/latest/?badge=latest)
+[![Documentation Status](https://readthedocs.org/projects/torchquad/badge/?version=main)](https://torchquad.readthedocs.io/en/main/?badge=main)
 
 
 
@@ -89,7 +89,7 @@ This is a brief guide for how to set up torchquad.
 ### Prerequisites
 
 We recommend using [conda](https://anaconda.org/conda-forge/torchquad), especially if you want to utilize the GPU. It will automatically set up CUDA and the cudatoolkit for you in that case.
-Note that torchquad also works on the CPU; however, it is optimized for GPU usage.
+Note that torchquad also works on the CPU; however, it is optimized for GPU usage. Currently torchquad only supports NVIDIA cards with CUDA. We are investigating future support AMD cards through [ROCm](https://pytorch.org/blog/pytorch-for-amd-rocm-platform-now-available-as-python-package/).
 
 For a detailed list of required packages, please refer to the [conda environment file](https://github.com/esa/torchquad/blob/main/environment.yml).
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -89,7 +89,7 @@ This is a brief guide for how to set up torchquad.
 ### Prerequisites
 
 We recommend using [conda](https://anaconda.org/conda-forge/torchquad), especially if you want to utilize the GPU. It will automatically set up CUDA and the cudatoolkit for you in that case.
-Note that torchquad also works on the CPU; however, it is optimized for GPU usage.
+Note that torchquad also works on the CPU; however, it is optimized for GPU usage. Currently torchquad only supports NVIDIA cards with CUDA. We are investigating future support AMD cards through [ROCm](https://pytorch.org/blog/pytorch-for-amd-rocm-platform-now-available-as-python-package/).
```
htoftevaag, line 92:
```suggestion
Note that torchquad also works on the CPU; however, it is optimized for GPU usage. Currently torchquad only supports NVIDIA cards with CUDA. We are investigating future support for AMD cards through [ROCm](https://pytorch.org/blog/pytorch-for-amd-rocm-platform-now-available-as-python-package/).
```

docs/source/install.rst
```diff
@@ -15,7 +15,8 @@ Prerequisites
 
 We recommend using `conda <https://docs.conda.io/en/latest/>`_, especially if you want to utilize the GPU. 
 It will automatically set up CUDA and the cudatoolkit for you in that case.
-Note that *torchquad* also works on the CPU; however, it is optimized for GPU usage.
+Note that *torchquad* also works on the CPU; however, it is optimized for GPU usage. 
+Currently torchquad only supports NVIDIA cards with CUDA. We are investigating future support AMD cards through [ROCm](https://pytorch.org/blog/pytorch-for-amd-rocm-platform-now-available-as-python-package/).
```
htoftevaag, line 19:
```suggestion
Currently torchquad only supports NVIDIA cards with CUDA. We are investigating future support for AMD cards through [ROCm](https://pytorch.org/blog/pytorch-for-amd-rocm-platform-now-available-as-python-package/).
```

htoftevaag (changes requested):
Added "for" two places (see suggestions)

htoftevaag (approved):

PR #92, opened 2021-07-16
Transition to PyTorch 1.9

htoftevaag:
Run deployment test


gomezzz (approved):
Looks, tested with tests and deployment test in a new env, no probs so far.

PR #93, opened 2021-07-16
Add image shields

htoftevaag:
Added square image shields to Readme. 

For Develop branch


gomezzz (changes requested):
It looks good but why do you want to go into develop? This only changes README, I think it should go directly into main?

gomezzz (approved):

PR #97, opened 2021-07-16
Deleted benchmark notebook

gomezzz:


htoftevaag (approved):
Deleted benchmark notebook OK

PR #106, opened 2021-08-06
Update rt d   docs fail

htoftevaag:
# Description

Summary of changes

* RtD fails due to cudatoolkit version

## How Has This Been Tested?

- Not tested yet, will run test


gomezzz (approved):
(Tried on my machine, works although conda still messes up installing pytorch_gpu . But that is unrelated, I think)

PR #113, opened 2021-08-11
Minor performance improvement in integration_grid.py

danielkelshaw:
# Description

Very minor change to `integration_grid.py`.

## Summary of changes

* By using `torch.ravel()` a view of the Tensor is used rather than copying it.
* It is also possible to stack with a dimension rather than transposing the Tensor after-the-fact.

## How Has This Been Tested?

- [x] Ran `pytest` in the `torchquad/tests` directory -- `7/7 passed`


gomezzz (approved):

gomezzz:
Thanks! :)

PR #115, opened 2021-08-12
Add complex numbers to docs

htoftevaag:
# Add complex numbers and Boole's to docs

Summary of changes:

* Added Boole's to tutorial
* Added complex numbers example to tutorial


docs/source/tutorial.rst
```diff
@@ -94,51 +96,58 @@ Now let’s get started! First, the general imports:
 
 
 
-One-dimensional integration
-----------------------------
+One-dimensional integration with complex numbers
```
gomezzz, line 99:
```suggestion
One-dimensional integration
```
the example needn't be complex so I would not make it seem as if the example is exclusively complex

docs/source/tutorial.rst
```diff
@@ -94,51 +96,58 @@ Now let’s get started! First, the general imports:
...
 to the `Trapezoid rule <https://www.math.ubc.ca/~pwalls/math-python/integration/trapezoid-rule/>`__
-and `Simpson’s rule <https://www.math.ubc.ca/~pwalls/math-python/integration/simpsons-rule/>`__
+and `Simpson's rule <https://www.math.ubc.ca/~pwalls/math-python/integration/simpsons-rule/>`__
 in one dimension.
 Similarly, `Tirthajyoti Sarkar <https://github.com/tirthajyoti>`__ has made a nice visual explanation of 
 `Monte Carlo integration in Python 
 <https://towardsdatascience.com/monte-carlo-integration-in-python-a71a209d277e>`__.
 
-Let ``f(x)`` be the function :math:`f(x) = e^{x} \cdot x^{2}`. Over the domain 
+The *torchquad* package also works with complex numbers. To showcase this, in the following example we
```
gomezzz, line 113:
```suggestion
*torchquad* is compatible with complex numbers. To showcase this, in the following example we 
```

docs/source/tutorial.rst
```diff
@@ -94,51 +96,58 @@ Now let’s get started! First, the general imports:
...
-and `Simpson’s rule <https://www.math.ubc.ca/~pwalls/math-python/integration/simpsons-rule/>`__
+and `Simpson's rule <https://www.math.ubc.ca/~pwalls/math-python/integration/simpsons-rule/>`__
 in one dimension.
 Similarly, `Tirthajyoti Sarkar <https://github.com/tirthajyoti>`__ has made a nice visual explanation of 
 `Monte Carlo integration in Python 
 <https://towardsdatascience.com/monte-carlo-integration-in-python-a71a209d277e>`__.
 
-Let ``f(x)`` be the function :math:`f(x) = e^{x} \cdot x^{2}`. Over the domain 
+The *torchquad* package also works with complex numbers. To showcase this, in the following example we 
+will let ``f(x)`` be the complex function :math:`f(x) = e^{3ix} \cdot x^{2}`. Over the domain
```
gomezzz, line 114:
```suggestion
will let ``f(x)`` be the complex function :math:`f(x) = e^{3ix} \cdot x^{2}` (real integrands can be used in the exact same way). Over the domain 
```

docs/source/tutorial.rst
```diff
@@ -94,51 +96,58 @@ Now let’s get started! First, the general imports:
...
+will let ``f(x)`` be the complex function :math:`f(x) = e^{3ix} \cdot x^{2}`. Over the domain 
 :math:`[0,2]`, the integral of ``f(x)`` is :math:`\int_{0}^{2} f(x) dx = 
-\int_{0}^{2} e^x \cdot x^2 dx = 2(e^{2} - 1) = 12.7781121978613004544...`
+\int_{0}^{2} e^{3ix} \cdot x^{2} dx = -\frac{2}{27} i ( 1 + (17 + 6i) e^{6i} )`, or, alternatively, 
+:math:`0.0748857930... - i1.4073621035...` 
+(see `Wolfram Alpha <https://www.wolframalpha.com/input/?i=integral+of+e%5E%283ix%29+*+x%5E%282%29+from+0+to+2>`__ 
+if you don't belive us).
 
-Let’s declare the function and a simple function to print the absolute error, 
+Let’s declare the ``f(x)`` function, a simple helper function to print the absolute error,
```
gomezzz, line 121:
```suggestion
Let’s declare the function ``f(x)``, a simple helper function to print the absolute error, 
```

docs/source/tutorial.rst
```diff
@@ -94,51 +96,58 @@ Now let’s get started! First, the general imports:
...
     def print_error(result,solution):
-        print("Results:",result.item())
+        print("Results:", result.item())
         print(f"Abs. Error: {(torch.abs(result - solution).item()):.8e}")
         print(f"Rel. Error: {(torch.abs((result - solution) / solution).item()):.8e}")
     
-    solution = 2*(torch.exp(torch.tensor([2.]))-1)
+    solution = -2.0j / 27.0 * (1.0 + (17.0 + 6.0j) * torch.exp(torch.tensor([6.0j])) )
 
 **Note that we are using the torch versions to ensure that all variables
```
gomezzz, line 136:
```suggestion
**Note that we are using the torch versions of functions like ``exp`` to ensure that all variables
```

docs/source/tutorial.rst
```diff
@@ -148,7 +157,8 @@ Let’s define the integration domain now and initialize the integrator - let’
 
 .. code:: ipython3
 
-    integration_domain = [[0, 2]] # Integration domain is always a list of lists to allow arbitrary dimensionality.
+    # Integration domain is always a list of lists to allow arbitrary dimensionality.
+    integration_domain = [[0, 2]]
```
gomezzz, line 161: This fails to show how a complex integration domain would function I think?

docs/source/tutorial.rst
```diff
@@ -178,10 +188,23 @@ Let’s see what type of value we get for different integrators.
...
+    boole = Boole()
+    result = boole.integrate(f, dim=1, N=101, integration_domain=integration_domain)
+    print_error(result,solution)
+
+
+.. parsed-literal::
+
+    **Output:** Results: (0.0748857855796814-1.4073619842529297j)
+            Abs. Error: 1.19441893e-07
+            Rel. Error: 8.47494519e-08
```
gomezzz, line 206: Here we should either:
- Reduce N to have differences between integrators
- Switch to double precision to have differences between integrators
- Mention that the numerical error is within range of floating point precision and hence no improvement can be seen for Boole

docs/source/tutorial.rst
```diff
@@ -206,15 +229,15 @@ Let’s see what type of value we get for different integrators.
 
 .. parsed-literal::
 
-    **Output:** Results: 21.83991813659668
-            Abs. Error: 9.06180573e+00
-            Rel. Error: 7.09166229e-01
+    **Output:** RuntimeError: result type ComplexFloat can't be cast to the desired output type Float
```
gomezzz, line 232: ??? This is very awkward.

docs/source/tutorial.rst
```diff
@@ -206,15 +229,15 @@ Let’s see what type of value we get for different integrators.
...
     
 
-Notably, Simpson’s method is already sufficient for a perfect solution here with 101 points. 
+Notably, Simpson's and Boole's methods are already really good here with only 101 points. 
+(You can try setting the function ``f(x) = torch.exp(x) * torch.pow(x,2)`` instead 
+to see Simpson's and Boole's methods find the perfect solution).
```
gomezzz, line 237: >  (You can try setting the function ``f(x) = torch.exp(x) * torch.pow(x,2)`` instead 
to see Simpson's and Boole's methods find the perfect solution).

I don't understand this? It's a harder function so the errors will be higher? So not perfect?

docs/source/tutorial.rst
```diff
@@ -303,15 +343,20 @@ We are working on giving the user more flexibility on this point.
...
 
-Note that the Monte Carlo methods are much more competitive for
-this case. The bad convergence properties of the trapezoid method are
-visible while Simpson’s rule is still OK given the comparatively smooth
-integrand.
+Note that the Monte Carlo methods are much more competitive in this case. 
+The bad convergence properties of the trapezoid method are visible while Simpson's 
+and Boole's rule are still OK given the comparatively smooth integrand. 
+
+Note that with the choice of sample points (3 per dimension), Boole's method cannot work.
```
gomezzz, line 350: I would change all to five then? Otherwise the comparison is pointless

Comparing VEGAS/MC with 59049 points to Boole with 9.765.625 is absurd

gomezzz (changes requested):

docs/source/tutorial.rst
htoftevaag: But I cannot do this on my computer 😭

docs/source/tutorial.rst
htoftevaag: No, we had 
`torch.exp(3.0j*x) * torch.pow(x,2)`
and then I suggest using
`torch.exp(x) * torch.pow(x,2)`

docs/source/tutorial.rst
htoftevaag: Would you rather hide it?

docs/source/tutorial.rst
gomezzz: or create a separate ticket for it to have another example with that

docs/source/tutorial.rst
gomezzz: yes don't run vegas for it if incompatible

docs/source/tutorial.rst
gomezzz: Ah, so it's not harder. However, e^x is not a polynomial, so it cannot be "perfectly" solved by either method. I would just remove this as it does not really add to informing the reader.

docs/source/tutorial.rst
gomezzz: Did it, see related ticket.

htoftevaag (commented):
I had some comments that Pablo and I discussed, see resolved comments

gomezzz (approved):

Issue #123, opened 2021-08-20
Allow support of unequal numbers of points per dimension


htoftevaag:
# Support for unequal numbers of points per dimension

## Desired Behavior / Functionality

Torchquad currently only supports equal numbers of points per dimension for the deterministic methods (that use integration_grid.py). The user might want to use more points in one dimension than the others (if, for instance, the integration domain in one dimension is much larger than in the others or the function is more complex in certain dimensions).
Exactly how this should be implemented can be a bit up the contributor - perhaps the ideal solution is allowing many different types of input (see below).

## What Needs to Be Done

Under `torchquad.integration_grid.py`, we have the following comment `# TODO Add that N can be different for each dimension`. 
Currently, what happens is that the number of sample points per dimension (`self._N`) is set to the _n_th root of the total number of sample points, _N_, where _n_ is the number of dimensions.
`self._N = int(N ** (1.0 / self._dim) + 1e-8)  # convert to points per dim ` (the 1e-8 term here is explained in the code).

It would be nice to allow `self_N` to be for instance `[16, 2, 2]` instead of `[4, 4, 4]`.

One solution could be to do the following:

If _N_ is an integer, then let `self._N` be the _n_th root of _N_.
If _N_ is a list/array, then let `self._N` be the list/array.

**Note:** For Monte Carlo and VEGAS, varying the number of points per dimension doesn't make sense.

## How Can It Be Tested

A good place to start is to check if the tests fail, in particular the integration_grid_test.py.
Ideally, additional tests should be added to integration_grid_test.py.

PR #133, opened 2021-12-16
Adaptive trapezoid

gomezzz:
# Description

Summary of changes

* Added adaptive_trapezoid
* Added adaptive_grid

## Outstanding work

- [x] Create some test functions to see if this is working (or use existing ones)
- [x] Debug the tests (currently errors are noticeably larger than for default trapezoid which does not make sense)
- [x] Optimize: There are several TODOs in the code for this (reusing computed points, allocating memory in a smarter way)
- [ ] Choose good baseline values for N_subdomains and initial N in each subdomain
- [ ] Add more logging (currently only debug in some places)
- [x] Create some code to visualize the created adaptive grids. This should help also with debugging.
- [ ] Improve refinement criterion
- [ ] Optimize for larger number of subdomains

There are a few larger topics still open then, but these probably deserve dedicated issues and PRs
- [x] Creating AdaptiveBoole and AdaptiveSimpson (should be fairly straight forward)
- Make the refinement not refine in all dimension equally, this is closely related to #123 
- Implement batched grid computations to allow larger N without having out of memory problems

Above points should help remedy some of the issues, right now performance on a simple integrand (`f = torch.sin(x[:,0] + (x[:,1]+1.)**2) * (x[:,2]+1.)`) is worse than for vanilla trapezoid, but that is somewhat expected.

image

image

## How Has This Been Tested?

- [ x ] Added new test


gomezzz:
Not working on for the moment.

PR #136, opened 2022-02-03
added surface plot to visualize adaptive grid and its function values

MrGweep:
Added function torchquad/plots/plot_adaptive_grid.py to visualize the adaptive grid with the corresponding function values on a 3D-scatter plot.

### DIFF
--- a/torchquad/plots/plot_adaptive_grid.py
+++ b/torchquad/plots/plot_adaptive_grid.py
@@ -15,11 +15,11 @@ def plot_grid(grid, dpi=100):
     fvals = None
     for subdomain in grid.subdomains:
         if points is None:
-            points = subdomain.points
-            fvals = subdomain.fval
+            points = subdomain.points.cpu().numpy()
+            fvals = subdomain.fval.cpu().numpy()
         else:    
-            points = np.concatenate([points,subdomain.points])
-            fvals = np.concatenate([fvals,subdomain.fval])
+            points = np.concatenate([points,subdomain.points.cpu().numpy()])
+            fvals = np.concatenate([fvals,subdomain.fval.cpu().numpy()])
     ax = fig.add_subplot(111, projection='3d')
     ax.plot_trisurf(points[:,0], points[:,1], fvals, color='white', edgecolors='grey', alpha=0.5)
     ax.scatter(points[:,0], points[:,1], fvals, c='red')
--- a/torchquad/plots/plot_adaptive_grid.py
+++ b/torchquad/plots/plot_adaptive_grid.py
@@ -2,7 +2,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 
 
-def plot_grid(grid, dpi=100):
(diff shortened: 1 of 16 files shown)


torchquad/plots/plot_adaptive_grid.py
```diff
@@ -0,0 +1,29 @@
+import matplotlib.pyplot as plt
+import numpy as np
+
+
+def plot_grid(grid, dpi=100):
```
gomezzz, line 5: For consistency, let's have function and filename match
```suggestion
def plot_adaptive_grid(grid, dpi=100):
```

gomezzz (changes requested):
I think, you also need to add the function to torchquad's `__init__.py` to expose it to outside users :)

gomezzz:
Ah, also we might want to expose the AdaptiveGrid of each integrator as well then, I think. Otherwise, users won't be able to access the grid. (we can just expose the grid through the integrator with a `get_grid()` in the AdaptiveNewtonCotes integrator in the PR #135 )

torchquad/plots/plot_adaptive_grid.py
```diff
@@ -0,0 +1,29 @@
...
+    fig = plt.figure(dpi=dpi)
+    points = None
+    fvals = None
+    for subdomain in grid.subdomains:
+        if points is None:
+            points = subdomain.points
+            fvals = subdomain.fval
+        else:    
+            points = np.concatenate([points,subdomain.points])
+            fvals = np.concatenate([fvals,subdomain.fval])
```
gomezzz, line 22:
```suggestion
        if points is None:
            points = subdomain.points.cpu().numpy()
            fvals = subdomain.fval.cpu().numpy()
        else:    
            points = np.concatenate([points,subdomain.points.cpu().numpy()])
            fvals = np.concatenate([fvals,subdomain.fval.cpu().numpy()])
```

Otherwise you get a
`TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.` if you use the GPU

gomezzz:
image

I think we might also want to adjust the pointsize based on N in the grid. With N = 10 000 it's already not interpretable otherwise

gomezzz:
Ah one more note. For high N it becomes very slow. We might consider disabling the trisurf part for N > 10000? That helps a lot with performance

torchquad/plots/plot_adaptive_grid.py
```diff
@@ -0,0 +1,29 @@
...
+
+    Args:
+        grid (AdaptiveGrid): AdaptiveGrid of evaluated function
+        dpi (int, optional): Plot dpi. Defaults to 100.
+    """
+
+    fig = plt.figure(dpi=dpi)
+    points = None
+    fvals = None
+    for subdomain in grid.subdomains:
```
gomezzz, line 16: I think this function only supports 2D grids, right? If so, we should raise an error if a non-2D grid is given

PR #139, opened 2022-03-04
Updated return types in docstirngs of .integrate functions according …

### LINKED ISSUE #138: Fix wrong return type in docstrings
Hi,
I would like to suggest that the docsrings of bascially all `.integrate` functions of the classes in `torchquad.integration` are updated as they are currently reading `Returns: float: Integral value`. However, the functions are actually returning an object of type `torch.Tensor`. Of course, this is neither a bug nor a big deal at all, but it results in the user getting warnings and complaints by the IDE/IntelliSense.<br>
If you agree that this would be a good idea / quality of life update then I can offer to do the job of creating a Pull Request.

By the way: `torchquad` is a cool package! Thanks for your work!

Best regards
Martin

MartinBubel:
Closes #138

# Description
The `.integrate` functions of classes `Boole`, `MonteCarlo`, `Simpson`, `Trapezoid`, and `VEGAS` have been reading `float`. However, they return a `torch.Tensor`. When using `torchquad` integrators in IDEs that use intellisense or comparable syntax highlighting tools, the user usually receives warnings from calling `torch.Tensor`-related operations such as `.backward()` or `.item()` on the `.integrate()` output, which by the syntax checker is expected to be a `float`.

Summary of changes
* Changed the return type in the docstring of the `.integrate` function of classes `Boole`, `MonteCarlo`, `Simpson`, `Trapezoid`, and `VEGAS` from `float` to `torch.Tensor`.

## Resolved Issues
- [ ] fixes #138 

## How Has This Been Tested?
Since no actual code was changed but only docstrings, the auto-tests should provide sufficient testing.

### DIFF
--- a/torchquad/integration/boole.py
+++ b/torchquad/integration/boole.py
@@ -24,7 +24,7 @@ class Boole(BaseIntegrator):
             integration_domain (list, optional): Integration domain, e.g. [[-1,1],[0,1]]. Defaults to [-1,1]^dim.
 
         Returns:
-            float: integral value
+            torch.Tensor: integral value
         """
 
         # If N is unspecified, set N to 5 points per dimension
--- a/torchquad/integration/monte_carlo.py
+++ b/torchquad/integration/monte_carlo.py
@@ -25,7 +25,7 @@ class MonteCarlo(BaseIntegrator):
             ValueError: If len(integration_domain) != dim
 
         Returns:
-            float: integral value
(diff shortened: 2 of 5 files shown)


gomezzz:
black formatting will be fixed in #137 (I think the action is failing due to a black update)

gomezzz (approved):
Thanks!

PR #141, opened 2022-04-21
Gaussian quadrature

### LINKED ISSUE #126: Add Gaussian quadrature methods
# Feature
Currently, torchquad only implements Newton-Cotes formulas as deterministic integration methods. A logical next step would be Gaussian quadrature methods.

## What Needs to Be Done
- [ ] Decide on integrators to implement (e.g. [Gauss-Legendre](https://en.wikipedia.org/wiki/Gauss%E2%80%93Legendre_quadrature)
- [ ] Implement integrator in integration folder, should inherit from BaseIntegrator
- [ ] Add docs for it
- [ ] Add tests for it, orient at Newton-Cotes tests

## How Can It Be Tested
With new tests!

elastufka:
# Add Gaussian quadrature methods

Base Gaussian quadrature class created

Several Gaussian quadrature methods implemented, for integration over [a,b], [0,inf] and [-inf,inf]

Minor changes to BaseIntegrator to allow for args for input functions

## Resolved Issues

- [X ] fixes #126 

## How Has This Been Tested?

- [X] See examples in docstrings

## How Has This _Not_ Been Tested?

All except Gauss-legendre have not been tested on the GPU. Most have also not been tested in multiple dimensions.

No backends besides Torch and numpy have been tested

## Known issues

Finding the roots of some polynomials (Jacobi, Laguerre) use scipy functions. I don't know if/how this is compatible with autoray.do()

Transforming points and weights for integration can be tricky with the GPU involved - need to be able to check/enforce that all tensors are on the same device

Gauss-Jacobi integration does not currently give the correct solution outside the interval [-1,1]

## Not yet implemented

other Gaussian methods including Gauss-Kronrod / QUADPACK


gomezzz:
I changed PR base as new features go into develop first :pray:

gomezzz:
Just request a review when I should review this!

elastufka:
Hi, 

I can't seem to be able to request a review through the usual method. However, please feel free to take a look at your convenience. 

All results for 1-dimensional integrals are correct now. You can see the tests that I did in this [Colab](https://colab.research.google.com/drive/1mggqd5mEeYAe-PCDxgzfjbCbhNd0JPpE?usp=sharing). 

Probably the integration does not work correctly for higher dimensions. When I first looked at this I was interested in vectorized integration - integrating over vectors of limits in one dimension - which is just the result of treating dims as the number of limits and not summing over all dimensions in the end. This is what quadpy does as well, and since it turns out that doing fixed-point and iterative integration in Torch is very slow even with large numbers of points, I won't be looking into it much more, or into high-dimension integration at all. (Also with the particular integral I wanted to do quickly, it looks like tensor-based backends won't help either since it is bogged down by a high-order cross-section calculation.)

Hopefully someone who has a better knowledge of how high-dimensional integration should work can quickly get it working based on the start that's been made here.

gomezzz:
(switched base to refresh diff)

gomezzz:
@elastufka First off, thank you for your efforts! The code looks very well structured and fits in neatly!

A few thoughts to integrate seemless with CI, current version etc.:

- [ ] Could you have a look at the merge conflicts / ideally merge current status of `develop` into this? 
- [ ] Thanks for the colab as well! We will need to add some unit tests for this to make sure the results stay correct etc. could you possibly add a test for the gaussian integrators similar to, e.g., [the trapezoid one](https://github.com/esa/torchquad/blob/main/torchquad/tests/trapezoid_test.py)? It can be very similar. Or you can use some of the results from your colab if you prefer.

> All except Gauss-legendre have not been tested on the GPU. Most have also not been tested in multiple dimensions.
> No backends besides Torch and numpy have been tested
I will try this out soon :)

> Finding the roots of some polynomials (Jacobi, Laguerre) use scipy functions. I don't know if/how this is compatible with autoray.do()

In general it is fine if not all integrators are compatible with everything as long as we throw meaningful exceptions / make sure we know. The easiest way for testing this would be

- [ ] Add the new integrators to GradientTests at https://github.com/esa/torchquad/blob/bbbb3782cda4ff56f0e8093102dadaf87517b2a0/torchquad/tests/gradient_test.py#L173
- [ ] Please add the version of scipy you used to the requirements [here](https://github.com/esa/torchquad/blob/main/setup.py), [here](https://github.com/esa/torchquad/blob/main/requirements.txt), [here](https://github.com/esa/torchquad/blob/main/environment.yml) and [here](https://github.com/esa/torchquad/blob/main/environment_all_backends.yml) (sorry for the redundancy, supporting CI, pip, conda and rtd has become a bit tedious). In my env with scipy 1.8.0 I currently get `module 'scipy' has no attribute 'special'` trying the gaussian integrators

> Transforming points and weights for integration can be tricky with the GPU involved - need to be able to check/enforce that all tensors are on the same device

Don't worry too much about this. I will try it out and we should see errors then. 

> Gauss-Jacobi integration does not currently give the correct solution outside the interval [-1,1]
- [ ] I would suggest to either thrown an error then for other bounds or transform the integrand into this automatically?

> Probably the integration does not work correctly for higher dimensions. When I first looked at this I was interested in vectorized integration - integrating over vectors of limits in one dimension - which is just the result of treating dims as the number of limits and not summing over all dimensions in the end. This is what quadpy does as well, and since it turns out that doing fixed-point and iterative integration in Torch is very slow even with large numbers of points, I won't be looking into it much more, or into high-dimension integration at all. (Also with the particular integral I wanted to do quickly, it looks like tensor-based backends won't help either since it is bogged down by a high-order cross-section calculation.)

Of course. We can always improve on this later. Ideally, could you maybe add an issue describing what needs improvements later and - if you have an idea already - how that could be done?

If this becomes too annoying / much time-investment for you I can also take care of some of the above things. I just want to make sure we stay future-proof. :)

Thanks!

gomezzz:
(I will wait for your reply before I dig in deeper)

elastufka:
* [X]  Could you have a look at the merge conflicts / ideally merge current status of `develop` into this?
 * [X]  Thanks for the colab as well! We will need to add some unit tests for this to make sure the results stay correct etc. could you possibly add a test for the gaussian integrators similar to, e.g., [the trapezoid one](https://github.com/esa/torchquad/blob/main/torchquad/tests/trapezoid_test.py)? It can be very similar. Or you can use some of the results from your colab if you prefer.
* [X]  Add the new integrators to GradientTests at https://github.com/esa/torchquad/blob/bbbb3782cda4ff56f0e8093102dadaf87517b2a0/torchquad/tests/gradient_test.py#L173

I added the tests but haven't run any of them yet. Pretty sure that even if the high-dimension tests pass, the results will be incorrect so that will need a closer look.

* [ ]  Please add the version of scipy you used to the requirements [here](https://github.com/esa/torchquad/blob/main/setup.py), [here](https://github.com/esa/torchquad/blob/main/requirements.txt), [here](https://github.com/esa/torchquad/blob/main/environment.yml) and [here](https://github.com/esa/torchquad/blob/main/environment_all_backends.yml) (sorry for the redundancy, supporting CI, pip, conda and rtd has become a bit tedious). In my env with scipy 1.8.0 I currently get `module 'scipy' has no attribute 'special'` trying the gaussian integrators

I didn't do the last one but looking at the scipy 1.6.0 documentation (minimum version according to setup.py), all the functions I use are already present there. The 'from scipy import special' line should be compatible with all versions

torchquad/integration/gaussian.py
```diff
@@ -0,0 +1,209 @@
...
+            wrapper_func (func,optional): function that performs any additional calculations required to get the proper points and weights, eg. for use in Gauss-Lobatto quadrature, or for transformation of interval in Gauss-Legendre quadrature. Default is None.
+        
+        Returns:
+            tuple(points, weights)
+            """
+        xi, wi = root_fn(*root_args) #can autoray work with scipy functions?
+        if isinstance(self._integration_domain,torch.Tensor):
+            device=self._integration_domain.device
+            xi=torch.from_numpy(xi).to(device)
+            wi=torch.from_numpy(wi).to(device)
```
gomezzz, line 38: is this a torch specific thing or should this also happen for other backends?

torchquad/integration/gaussian.py
```diff
@@ -0,0 +1,209 @@
...
+        
+        Args:
+            root_fn (func): function to use for computing sample points and weights via the roots of appropriate polynomials
+            root_args (tuple): arguments required for root-finding function, most commonly the degree N
+            wrapper_func (func,optional): function that performs any additional calculations required to get the proper points and weights, eg. for use in Gauss-Lobatto quadrature, or for transformation of interval in Gauss-Legendre quadrature. Default is None.
+        
+        Returns:
+            tuple(points, weights)
+            """
+        xi, wi = root_fn(*root_args) #can autoray work with scipy functions?
```
gomezzz, line 34: > #can autoray work with scipy functions?
were you able to settle this question?

torchquad/integration/gaussian.py
```diff
@@ -0,0 +1,209 @@
...
+#    def wrapper_func(self,xi,wi): #sould override default None
+
+
+#class GaussRadau(Gaussian):
+#    """Gauss Radau quadrature rule in torch. See """
+#
+#    def __init__(self):
+#        super().__init__()
+#        self.name="Gauss-Lobatto"
+#        self.root_fn=scipy.special.roots_jacobi #need additional function to operate on xi,wi to get the correct weights and points from the Jacobi polynomial roots...
```
gomezzz, line 208: Maybe we can describe the missing thing with this code snippet in the associated issue in case we want to ad dthis later? Just to not have dead code in there

torchquad/integration/gaussian.py
```diff
@@ -0,0 +1,209 @@
...
+        super().__init__()
+        
+    def wrapper_func(self,xi,wi):  #scale from [-1,1] to [a,b]
+        a,b=self._integration_domain.T
+        x0 = 0.5 * (1.0 - xi) #self.points)
+        x1 = 0.5 * (1.0 + xi)#self.points)
+        xi = anp.outer(a, x0) + anp.outer(b, x1)
+        return xi,wi
+        
+    def _get_constant_multiplier(self): #do this somewhere else? with points and weights maybe?
```
gomezzz, line 113: could you add a docstring to this method? also the name seems very cryptic, if another less generic name would be possible would be great too

gomezzz (changes requested):
Mostly minor comments, could you please check the failures on the github actions? It seems there are syntax and linting errors, so currently the unit tests are not runnin. See https://github.com/esa/torchquad/runs/6457938281?check_suite_focus=true and https://github.com/esa/torchquad/runs/6457938431?check_suite_focus=true (you can see these, right?). Also please autoformat once with black :)

ilan-gold:
woops didn't mean to push here....but also the PR has been untouched and also did not appear to work...

ilan-gold:
Ok so I'm definitely making progress here.  All of the 1d integrand tests (at least for numpy) pass now.

ilan-gold:
Now need to the multi-dimensional integrand case, resizing the weights seems like all.  maybe an `einsum`

ilan-gold:
Ok all tests pass now for Gauss-Legendre integration.  Not sure what we want to do about the other ones, all of which involve some sort of restriction that the current full suite of tests does not obey (domain or type of function).  So we'll need to do one of the following:
(a) make new testing files
(b) don't merge those integrals (they are after all fairly trivial to implement and extremely restricted)
(c) something inbetween

ilan-gold:
Also, the multi-dimensional integrand case is supported, which brings full parity with `fixed_quad` (my original interest after all: https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.fixed_quad.html).  I would finish #160 first since it is merged into this branch and then we can move forward with this one.

ilan-gold:
It looks like this will be at a minimum 5x faster than scipy for reasonably sized systems for me (say ~50x150 at `N=60`), and caps out at least 40x for massive systems (will need to check this).  So this is huge for me.

gomezzz:
@ilan-gold wow, you have been busy! :) 

> Ok all tests pass now for Gauss-Legendre integration. Not sure what we want to do about the other ones, all of which involve some sort of restriction that the current full suite of tests does not obey (domain or type of function). So we'll need to do one of the following: (a) make new testing files (b) don't merge those integrals (they are after all fairly trivial to implement and extremely restricted) (c) something inbetween

Can you elaborate a bit what the problem is, I am not sure I get it?

ilan-gold:
Added JIT and ensured gradients are actually preserved!

gomezzz:
> > @ilan-gold wow, you have been busy! :)
> > > Ok all tests pass now for Gauss-Legendre integration. Not sure what we want to do about the other ones, all of which involve some sort of restriction that the current full suite of tests does not obey (domain or type of function). So we'll need to do one of the following: (a) make new testing files (b) don't merge those integrals (they are after all fairly trivial to implement and extremely restricted) (c) something inbetween
> > 
> > 
> > Can you elaborate a bit what the problem is, I am not sure I get it?
> 
> So the other integrators all have restrictions on how you can use them that the current test suite does not obey, like domain restrictions or integrand restrictions (they must be of a certain form). So we'd have to write new tests just for them. I personally do not think it's worth it.

Ah, you mean Gauss-Jacobi, Gauss-Hermite etc? Ok for me to leave that out, I think

ilan-gold:
Yup exactly those.  I think it could be a nice demo in the docs

ilan-gold:
@gomezzz I have not forgotten about this!  I just finished my exam so I am circling back to this.  I will put this in a stable place by the end of next week if not sooner!

gomezzz:
> @gomezzz I have not forgotten about this! I just finished my exam so I am circling back to this. I will put this in a stable place by the end of next week if not sooner!

@ilan-gold Great to hear! Hope the exams went well! :) Take your time, no rush. I will look at creating a release after that. :v: 

Btw. if you are interested in continuing with torchquad and joining me as a maintainer, I would be very open to that! :)

ilan-gold:
@gomezzz so I've run into a bit of an issue - I think that JIT is broken here for JAX.  Do you have any pointers/advice on getting JAX and autoray to work well together in this case? I'm trying to get this part working:

```python
File ~/torchquad/torchquad/integration/gaussian.py:59, in Gaussian._weights(self, N, dim, backend, requires_grad)
     52     return anp.prod(
     53         anp.stack(
     54             list(anp.meshgrid(*([weights] * dim))), like=backend, dim=0
     55         ),
     56         axis=0,
     57     ).ravel()
     58 else:
---> 59     return anp.prod(
     60         anp.meshgrid(*([weights] * dim), like=backend), axis=0
     61     ).ravel()

TracerArrayConversionError: The numpy.ndarray conversion method __array__() was called on the JAX Tracer object Traced<ShapedArray(float32[3,3,3])>with<DynamicJaxprTrace(level=0/1)>
The error occurred while tracing the function _weights at /home/ig62/torchquad/torchquad/integration/gaussian.py:37 for jit. This value became a tracer due to JAX operations on these lines:

  operation a:f32[3] = convert_element_type[new_dtype=float32 weak_type=False] b
    from line /home/ig62/masters-thesis/venv/lib/python3.8/site-packages/autoray/autoray.py:79 (do)
See https://jax.readthedocs.io/en/latest/errors.html#jax.errors.TracerArrayConversionError
```

I added `jit` decorators to non-static functions for jax - it seemed like `torch` worked fine but maybe I wasn't actually jit'ing it.  Thanks!

ilan-gold:
I think the problem goes beyond this PR - I think `jax` broke with the last PR because of the changes to `calculate_result`.  I think it should probably be refactored.  I can make a PR into this one to fix this.

gomezzz:
> I think the problem goes beyond this PR - I think `jax` broke with the last PR because of the changes to `calculate_result`. I think it should probably be refactored. I can make a PR into this one to fix this.

Running test CI again to confirm, will update in a sec :v:

gomezzz:
On the branch for this PR test seems to be passing but have a lot of warnings 🤔 

```
=============================== warnings summary ===============================
torchquad/tests/boole_test.py: 4 warnings
torchquad/tests/gauss_test.py: 4 warnings
torchquad/tests/gradient_test.py: 3 warnings
torchquad/tests/integrator_types_test.py: 4 warnings
torchquad/tests/monte_carlo_test.py: 4 warnings
torchquad/tests/simpson_test.py: 7 warnings
torchquad/tests/trapezoid_test.py: 4 warnings
  /home/runner/work/torchquad/torchquad/torchquad/tests/../integration/utils.py:255: UserWarning: DEPRECATION WARNING: In future versions of torchquad, an array-like object will be returned.
    warnings.warn(

torchquad/tests/boole_test.py::test_integrate_torch
  /home/runner/micromamba-root/envs/torchquad/lib/python3.10/site-packages/torch/functional.py:504: UserWarning: torch.meshgrid: in an upcoming release, it will be required to pass the indexing argument. (Triggered internally at /home/conda/feedstock_root/build_artifacts/pytorch-recipe_1675740247391/work/aten/src/ATen/native/TensorShape.cpp:3190.)
    return _VF.meshgrid(tensors, **kwargs)  # type: ignore[attr-defined]

torchquad/tests/gauss_test.py::test_integrate_torch
  /home/runner/micromamba-root/envs/torchquad/lib/python3.10/site-packages/autoray/autoray.py:79: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
    return get_lib_fn(backend, fn)(*args, **kwargs)

torchquad/tests/monte_carlo_test.py::test_integrate_jax
  /home/runner/micromamba-root/envs/torchquad/lib/python3.10/site-packages/autoray/autoray.py:79: UserWarning: Explicitly requested dtype <class 'jax.numpy.float64'> requested in array is not available, and will be truncated to dtype float32. To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See https://github.com/google/jax#current-gotchas for more.
    return get_lib_fn(backend, fn)(*args, **kwargs)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================= 52 passed, 33 warnings in 106.23s (0:01:46) ==================
```

Also pass on `develop` albeit with a bit fewer but still a lot of warnings

```
 =============================== warnings summary ===============================
torchquad/tests/boole_test.py: 4 warnings
torchquad/tests/gradient_test.py: 3 warnings
torchquad/tests/integrator_types_test.py: 4 warnings
torchquad/tests/monte_carlo_test.py: 4 warnings
torchquad/tests/simpson_test.py: 7 warnings
torchquad/tests/trapezoid_test.py: 4 warnings
  /home/runner/work/torchquad/torchquad/torchquad/tests/../integration/utils.py:255: UserWarning: DEPRECATION WARNING: In future versions of torchquad, an array-like object will be returned.
    warnings.warn(

torchquad/tests/boole_test.py::test_integrate_torch
  /home/runner/micromamba-root/envs/torchquad/lib/python3.10/site-packages/torch/functional.py:504: UserWarning: torch.meshgrid: in an upcoming release, it will be required to pass the indexing argument. (Triggered internally at /home/conda/feedstock_root/build_artifacts/pytorch-recipe_1675740247391/work/aten/src/ATen/native/TensorShape.cpp:3190.)
    return _VF.meshgrid(tensors, **kwargs)  # type: ignore[attr-defined]

torchquad/tests/monte_carlo_test.py::test_integrate_jax
  /home/runner/micromamba-root/envs/torchquad/lib/python3.10/site-packages/autoray/autoray.py:79: UserWarning: Explicitly requested dtype <class 'jax.numpy.float64'> requested in array is not available, and will be truncated to dtype float32. To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See https://github.com/google/jax#current-gotchas for more.
    return get_lib_fn(backend, fn)(*args, **kwargs)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================== 48 passed, 28 warnings in 84.06s (0:01:24) ==================
```

Are the warnings related to your problems, @ilan-gold ? I created a separate issue for them #163 

Or is a there a separate minimal example for the error you posted? Then we can create another issue for that. :)

ilan-gold:
@gomezzz Oh my gosh I'm so sorry, I forgot the word `jit`, which we don't test.

ilan-gold:
Wait no, I didn't forget it.  But we don't test `jit`, so this makes sense.  Things should be passing.

gomezzz:
I just looked a bit back in time, before #160 we had

```=================== 48 passed, 1 warning in 68.67s (0:01:08) ===================``` ( https://github.com/esa/torchquad/actions/runs/3830307450/jobs/6606968775 )

and then 

``` ================== 48 passed, 28 warnings in 73.58s (0:01:13) ==================``` ( https://github.com/esa/torchquad/actions/runs/3995901069/jobs/6855308819 )

So at the least the warnings seems to be related. Seems I should have complained more :D Sorry I only looked for the ✅ .... 🙈

**Comment by gomezzz** [comment]
(as an afterthought, this could also be related to some updates in other packages, doesn't necessarily have to be in #160 , I noticed e.g. autoray got some updates recently.)

**Comment by ilan-gold** [comment]
Ok `jit` stopped working because of the decorator.  I think the answer is to just not use `*args` and `**kwargs`

**Comment by gomezzz** [comment]
> Wait no, I didn't forget it. But we don't test `jit`, so this makes sense. Things should be passing.

That is suboptimal :D I added an issue ( #164 ) to add tests for it :v:

**Comment by ilan-gold** [comment]
Ok I fixed it - should be pretty straightforward!

**Comment by ilan-gold** [comment]
@gomezzz I opened #165 - this should fix the problem.  Separately I can add `jit` tests.

**Comment by ilan-gold** [comment]
Yay!

**Comment by gomezzz** [comment]
> I do not know what is going on with that test.

I had this problem in another repo as well, it's only the posting of the coverage comment, not sure if the thing I implemented is quite as robust as I had hoped :smiling_face_with_tear: Will investigate soon if it continues. In any event not related to this PR, I think.

File: torchquad/__init__.py [hunk 5]
```diff
@@ -22,16 +25,15 @@
 from .utils.deployment_test import _deployment_test
 
 __all__ = [
+    "GridIntegrator",
```
gomezzz, line 28: Do we want to expose these to the users? :thinking: 

Or is this with the idea of them also defining their own integrators based on them? Then it could be nice but we might want to either give an example or at least mention them in the docs :thinking: (if you prefer we can create a new issue for that)

torchquad/integration/base_integrator.py
```diff
@@ -77,8 +79,15 @@ def evaluate_integrand(fn, points, weights=None,args=None):
             )
 
         if weights is not None:
+            if len(result.shape) > 1:
```
gomezzz, line 82: Could you add a comment here explain this line? I think this is related to the changes from #160 ?

torchquad/integration/gaussian.py
```diff
@@ -1,99 +1,140 @@
...
-        logger.info(f"Computed integral was {integral}.")
-
-        return integral
+        roots = anp.array(self._cached_points_and_weights(N)[0], like=backend)
+        if requires_grad:
+            roots.requires_grad = True
+        return roots
+
+    @property
+    def _grid_func(self):
```
gomezzz, line 76: missing docstring :)

torchquad/integration/gaussian.py
```diff
@@ -1,99 +1,140 @@
...
+
+        Args:
+            a (backend tensor): lower bound
+            b (backend tensor): upper bound
+            roots (backend tensor): polynomial nodes
+
+        Returns:
+            backend tensor: rescaled roots
+        """
+        return roots
```
gomezzz, line 93: ? not sure if intentional? :)

torchquad/integration/grid_integrator.py
```diff
@@ -0,0 +1,268 @@
...
+
+    @property
+    def _grid_func(self):
+        def f(a, b, N, requires_grad=False, backend=None):
+            return _linspace_with_grads(a, b, N, requires_grad=requires_grad)
+
+        return f
+
+    def _weights(self, N, dim, backend, requires_grad=False):
+        return None
```
gomezzz, line 27: This should never be called, right? If so, then we might as well raise an exception in case it somehow does?

torchquad/integration/grid_integrator.py
```diff
@@ -0,0 +1,268 @@
...
+            backend tensor: Quadrature result
+        """
+        # Reshape the output to be [integrand_dim,N,N,...] points instead of [integrand_dim,dim*N] points
+        integrand_shape = function_values.shape[1:]
+        dim_shape = [n_per_dim] * dim
+        new_shape = [*integrand_shape, *dim_shape]
+        # We need to use einsum instead of just reshape here because reshape does not move the axis - it only reshapes.
+        # So the first line generates a character string for einsum, followed by moving the first dimension i.e `dim*N`
+        # to the end.  Finally we reshape the resulting object so that instead of the last dimension being `dim*N`, it is
+        # `N,N,...` as desired.
```
gomezzz, line 76: Could you add to the comment what chr(i+65) is? :)

gomezzz (changes requested):
A bunch of mostly minor comments :)

torchquad/tests/gauss_test.py
```diff
@@ -84,22 +74,72 @@ def _run_gaussian_tests(backend, _precision):
 
     # 10D Tests
     N = (60**3) * 3
```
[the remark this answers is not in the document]
ilan-gold: Yeah, this was a guess.  I'm not really sure what we should be going for here.

torchquad/integration/gaussian.py
ilan-gold: Yes, I just put some default behavior here i.e., for integrals over a fixed interval that don't need rescaling as is the case with some of the other Gaussian integrals.

torchquad/integration/grid_integrator.py
ilan-gold: I made this an implemented method in the base class because the Gaussian integral inherits from this and I've implemented everything here around generating the grid etc. which requires this method.

ilan-gold:
@gomezzz I generally cleaned up the tests.  I also addressed the other comments and left the ones unresolved where we might have a discussion.

torchquad/tests/gauss_test.py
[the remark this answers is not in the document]
gomezzz: N_d points per dim would lead to (N_d - 1) * (degree+1) points in that dimension, right? Then I would propose to maybe go for 3**10 or something like that? For Boole, e.g. , by that logic we picked 5**10 but that is already quite some points :upside_down_face:

torchquad/integration/grid_integrator.py
gomezzz: Sure, I am just thinking practically speaking this function should always be overwritten by any derived class, right? And if not, then it at least shouldn't be called. So instead of returning None and potentially failing silently if either it wasn't overwritten or called by accident we should raise an exception?

gomezzz:
@ilan-gold I am trying to fix the CI in #169 but on that PR it works. Could you try to merge the branch into this one to see if I fixed it? Otherwise, my suspicion is that it maybe because the PR comes from a fork? There seems to be some permission problems as described here https://github.com/MishaKav/pytest-coverage-comment/issues/68 (the mentioned repo setting is, however, correct, but maybe the one in the fork also affects this? Not quite sure yet what is going on :) )

ilan-gold:
A class MC failure @gomezzz - I will re-run but posting for posterity: https://github.com/esa/torchquad/actions/runs/4530799642/jobs/7980177519?pr=141

Not really sure what to do about these MC failures in the single-dimension integral domain case.  It seems like they are generally against the concept of unit-testing (since they are random).  Maybe we could set a seed?

gomezzz:
> Not really sure what to do about these MC failures in the single-dimension integral domain case. It seems like they are generally against the concept of unit-testing (since they are random). Maybe we could set a seed?

I think they are already seeded, but not all the frameworks provide deterministic seeding on different machines, iirc? I am afraid we may just need to increase the bounds instead :shrug:

ilan-gold:
@gomezzz I increased the bounds.  I think everything is passing now.  I have one more small PR based on https://github.com/ilan-gold/torchquad/tree/special_grid to make - basically allowing the user to disable a check and then generalizing the usage of `integration_domain` in a few places to use it directly instead of the two separate bounds i.e., `[a,b]` instead of passing `a` and `b`

torchquad/integration/grid_integrator.py
ilan-gold: The idea here is that some integration methods have weights and some do not, but the `GridIntegrator` should be able to handle both.  So if you prefer I can make the base class method abstract here instead of having this default behavior and then add this implementation to the `NewtonCotes` class instead tor return `None`.  Let me know :)

ilan-gold:
don't know what to say here @gomezzz about this coverage business...

torchquad/integration/grid_integrator.py
gomezzz: > The idea here is that some integration methods have weights and some do not, but the `GridIntegrator` should be able to handle both. So if you prefer I can make the base class method abstract here instead of having this default behavior and then add this implementation to the `NewtonCotes` class instead tor return `None`. Let me know :)

Yes, let's do that then :)

PR #154, opened 2022-06-21
Change the backend argument behaviour

### LINKED ISSUE #153: Change behavior of 'backend'
# Feature
After upgrading from 0.2.3 to 0.3.0 I encountered an issue where the backend grid had changed from a pytorch tensor to a numpy array. After making a cursory read of the documentation, I attempted setting the `backend` flags both in the integrator and using `set_backend()`. After this did not work, I was about to file an issue when I looked more closely and found that this input was ignored when the `integration_bounds` variable was not a tensor. I was then able to fix my issue. However, I think this decision should be reconsidered, as I did not find it intuitive.

`backend` should be less ignored - at the very least, a warning should be raised stating that it has been ignored in favor of the integration bounds.

FHof:
# Description

This affects situations where a user explicitly specifies the `backend` argument and uses a tensor for `integration_domain` at the same time.
* If the `integration_domain`'s backend does not correspond to the `backend` argument value, prefer the `backend` argument and show a warning.
  In this case the `integration_domain`'s dtype is ignored, so a globally configured dtype is used instead.
* If the `integration_domain`'s backend corresponds to the `backend` argument value, ignore the `backend` argument.
  In this case the `integration_domain`'s dtype is used and a globally configured dtype is ignored.

## Resolved Issues

-  fixes #153

## How Has This Been Tested?

-  Additional tests in `utils_integration_test.py`

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ Note also that installing PyTorch with *pip* may **not** set it up with CUDA sup
 Here are installation instructions for other numerical backends:
    ```sh
    conda install "tensorflow>=2.6.0=cuda*" -c conda-forge
-   pip install "jax[cuda]>=0.2.22" --find-links https://storage.googleapis.com/jax-releases/jax_releases.html # linux only
+   pip install "jax[cuda]>=0.2.22" --find-links https://storage.googleapis.com/jax-releases/jax_cuda_releases.html # linux only
    conda install "numpy>=1.19.5" -c conda-forge
    ```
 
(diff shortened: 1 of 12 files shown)


FHof:
The JAX developers have recently moved the wheel files for `jaxlib` with CUDA to a different webpage: https://github.com/google/jax/commit/e780a402e8e648deb3185442bf7c3701e44cc2c9. Perhaps it works again with the new installation instructions.

There also is a very recent issue about an incompatibility with the new NumPy version, which might cause problems: https://github.com/google/jax/issues/11221

gomezzz (approved):
:+1: thanks! :)

PR #160, opened 2023-01-03
(feat): multi-dimensional integrand

ilan-gold:
# Description
There might be a better word to describe this, but I figured this was descriptive to distinguish from integration dimensions.

Summary of changes

- Allow for an integrand which is multi-dimensional for all integration methods (except VEGAS for now) i.e a matrix/vector of values

## Resolved Issues

- [x] fixes #155

## How Has This Been Tested?

- Tests have been added to ensure that the feature works over simple 2x1 ("in both senses" i.e `(2,1)` and `(2,)`), 2x2, and 3x3 dimensional integrands 

Marking this as draft while we resolve some questions, namely how badly we want VEGAS (I would need to read the paper, which I am fine doing if y'all do not want to accept the PR without a full suite of features)

PR #166, opened 2023-03-06
(feat): add JIT integrator tests

### LINKED ISSUE #164: Add tests for JIT
# Feature
[Seems we are not testing JIT functionality](https://github.com/esa/torchquad/pull/141#issuecomment-1455757843), we should find a way to test it.

## What Needs to Be Done
- Test it :) , JIT can be used as described [here](https://torchquad.readthedocs.io/en/main/tutorial.html#compiling-the-integrate-method)

## How Can It Be Tested
🥁 with the tests

ilan-gold:
# Description

Adds tests for JIT integration

## Resolved Issues

- [x] fixes #164 

## How Has This Been Tested?

- [x] Added 1d JIT tests

## Related Pull Requests

- #PR

### DIFF
--- a/torchquad/tests/boole_test.py
+++ b/torchquad/tests/boole_test.py
@@ -64,10 +64,11 @@ def _run_boole_tests(backend, _precision):
     # JIT Tests
     if backend != "numpy":
         N = 401
-
-        integrate = bl.get_jit_compiled_integrate(
-            dim=1, N=N, backend=backend
-        )
+        def integrate(*args, **kwargs):
+            jit_integrate = bl.get_jit_compiled_integrate(
+                dim=1, N=N, backend=backend
+            )
+            return jit_integrate(*args, **kwargs)
         errors, funcs = compute_integration_test_errors(
             integrate,
             {},
--- a/torchquad/tests/boole_test.py
+++ b/torchquad/tests/boole_test.py
@@ -61,6 +61,27 @@ def _run_boole_tests(backend, _precision):
     # for error in errors:
     # assert error < 5e-9
 
+    # JIT Tests
+    if backend != "numpy":
+        N = 401
+
+        integrate = bl.get_jit_compiled_integrate(
+            dim=1, N=N, backend=backend
+        )
+        errors, funcs = compute_integration_test_errors(
+            integrate,
+            {},
+            integration_dim=1,
+            use_complex=True,
+            backend=backend,
+        )
+        print(f"1D Boole JIT Test passed. N: {N}, backend: {backend}, Errors: {errors}")
+        # Polynomials up to degree 5 can be integrated almost exactly with Boole.
+        for err, test_function in zip(errors, funcs):
+            assert test_function.get_order() > 5 or err < 6.33e-11
+        for error in errors:
+            assert error < 6.33e-11
+
 
 test_integrate_numpy = setup_test_for_backend(_run_boole_tests, "numpy", "float64")
 test_integrate_torch = setup_test_for_backend(_run_boole_tests, "torch", "float64")
--- a/torchquad/tests/boole_test.py
+++ b/torchquad/tests/boole_test.py
@@ -67,6 +67,8 @@ def _run_boole_tests(backend, _precision):
         jit_integrate = None
 
         def integrate(*args, **kwargs):
+            # this function initializes the jit_integrate variable with a jit'ed integrate function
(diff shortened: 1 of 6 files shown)


ilan-gold:
@gomezzz I've added some tests for this that work.  There are a few catches.  The main thing is that we don't test repeated integration for the same integrand-N-dim JIT combination (which is the point of JIT, but maybe not something we actually need to test).  This is because the test suite doesn't really track `integrand_shape` in the right places besides whether or not the integrand is multi-dimensional in general (from #160).  So we have a few options here:

1.  Refactor the test suite to allow for finer grained control over what is run and how (i.e., more explicitly tracking `integrand_shape`) - so a different PR in addition to this one.
2. Add another step to this PR to allow for re-running the 1d tests with a single JIT call (the functionality for which already exists), as opposed to this PR where JIT happens every time.
3. Keep what we have (i.e., JIT every integrand).

ilan-gold:
P.S One advantage of refactoring would be that we don't have to run all the tests.  I'm not sure what the goal of "unit testing" JIT is besides functionality i.e., I'm not sure we actually need to check accuracy of every single integral.  We're more interested in not-crashing.

gomezzz:
> P.S One advantage of refactoring would be that we don't have to run all the tests. I'm not sure what the goal of "unit testing" JIT is besides functionality i.e., I'm not sure we actually need to check accuracy of every single integral. We're more interested in not-crashing.

Agree :)

> So we have a few options here:

Any of them are okay for me. At the moment the tests take ~4min which is fine and I'll take any improvement :D

torchquad/tests/boole_test.py
```diff
@@ -61,6 +61,65 @@ def _run_boole_tests(backend, _precision):
...
     # assert error < 5e-9
 
+    # JIT Tests
+    if backend != "numpy":
+        N = 401
+        jit_integrate = None
+
+        def integrate(*args, **kwargs):
+            nonlocal jit_integrate
+            if jit_integrate is None:
```
gomezzz, line 71: Could you add a small comment explaining the though process here? It is a bit confusing :thinking:

torchquad/tests/boole_test.py
```diff
@@ -61,6 +61,65 @@ def _run_boole_tests(backend, _precision):
...
+            None  # set to None again so can be re-used with new integrand shape
+        )
+
+        def integrate(*args, **kwargs):
+            nonlocal jit_integrate
+            if jit_integrate is None:
+                jit_integrate = bl.get_jit_compiled_integrate(
+                    dim=1, N=N, backend=backend
+                )
+            return jit_integrate(*args, **kwargs)
```
gomezzz, line 104: is the redeclaration necessary? was declared just above? :thinking:

gomezzz: Same two comments apply to some of the other files as well, I think :v:

torchquad/tests/simpson_test.py
```diff
@@ -89,6 +89,65 @@ def _run_simpson_tests(backend, _precision):
     for error in errors:
         assert error < 5e-9
 
+    if backend != "numpy":
```
gomezzz, line 92: comment missing that the JIT section starts here (you had one in the others)

torchquad/tests/trapezoid_test.py
```diff
@@ -86,6 +86,65 @@ def _run_trapezoid_tests(backend, _precision):
     for error in errors:
         assert error < 7000
 
+    if backend != "numpy":
```
gomezzz, line 89: comment missing that the JIT section starts here (you had one in the others)

ilan-gold: done!

torchquad/tests/simpson_test.py
ilan-gold: done!

torchquad/tests/boole_test.py
ilan-gold: done (i think)

ilan-gold:
Well that now-failing test is flaky - fails sometimes, passes others.

gomezzz:
> Well that now-failing test is flaky - fails sometimes, passes others.

let's increase the bounds on this one? 🤔

ilan-gold:
Sounds good to me

gomezzz (approved):
👍

PR #173, opened 2023-04-21
(feat): add customization API for `IntegrationGrid`

ilan-gold:
# Description
Focused on customizing the integration domain, this PR allows the user to turn off checks for the domain and also gives a bit more freedom to the `grid_func` to not have to deal only with `a` and `b` in `[a,b]` but rather just takes is `[a,b]`.

Summary of changes

* add `disable_integration_domain_check` argument where applicable
* change usage of `a` and `b` to just `integration_domain`

## How Has This Been Tested?

- The existing suite of tests should suffice by not breaking since there is not new feature really here.


gomezzz:
> # Description
> 
> Focused on customizing the integration domain, this PR allows the user to turn off checks for the domain and also gives a bit more freedom to the `grid_func` to not have to deal only with `a` and `b` in `[a,b]` but rather just takes is `[a,b]`.
> 
> Summary of changes
> 
>     * add `disable_integration_domain_check` argument where applicable
> 
>     * change usage of `a` and `b` to just `integration_domain`
> 
> 
> ## How Has This Been Tested?
> 
>     * The existing suite of tests should suffice by not breaking since there is not new feature really here.

@ilan-gold So, to make sure I remember correctly: This allows integrating multiple integrands on different domains by transforming each integral into [0,1], right?

If so:
* Does it work for the other integrators (Newton Cotes, MC)? Should it?
* Can we add a tiny example somewhere how this can be used? I don't know how to test it 🥲 
* Should we maybe change the domain in one of the test cases to use this, otherwise it isn't really tested, is it?

torchquad/integration/utils.py
```diff
@@ -177,7 +177,7 @@ def _check_integration_domain(integration_domain):
                     integration_domain,
                     " does not specify a valid integration bound.",
                 )
-            if bounds[0] > bounds[1]:
+            if anp.all(bounds[0] > bounds[1]):
```
gomezzz, line 180:
```suggestion
            if anp.any(bounds[0] > bounds[1]):
```
Shouldn't this be this way since we want to raise an exception if any domain dimension is wrong?

ilan-gold: > @ilan-gold So, to make sure I remember correctly: This allows integrating multiple integrands on different domains by transforming each integral into [0,1], right?

Kind of - it allows process the same integrand repeatedly over a grid (as per my previous work) but each part of the grid gets its own custom `[a,b]` domain.

>     * Does it work for the other integrators (Newton Cotes, MC)? Should it?

You have to write a custom integrator.  I did.

>     * Can we add a tiny example somewhere how this can be used? I don't know how to test it 🥲

The "tiny example" I have is 60 lines ATM.  We could probably bring that down to 50 if we used a non-gaussian integrator (I need about 10 lines for resizing the roots).  I think my issue with adding a small example is that it then becomes something that also needs to be "correct."  

>     * Should we maybe change the domain in one of the test cases to use this, otherwise it isn't really tested, is it?

We can do this to check for the exception, at least.  I think this would actually count for testing the thing - with a "different" domain, either it fails with the check (good) or it fails in some horrible way later without the check (also a good sign).  So I'm not sure we need a small example to test this.  But perhaps we could just have a small mock example that makes the check or doesn't, and then checks the resulting integration domain (or something like that) so we don't have to worry about correctness of the integrand evaluation etc.  i.e., isolating this.

ilan-gold: So to sum up, I think a "tiny example" could be a mock fixture that doesn't even do integration but just wraps the `calculate_grid` method.  We could also just test the `IntegrationGrid` directly using the existing fixture.  Whichever you prefer.

torchquad/integration/integration_grid.py
```diff
@@ -23,15 +25,24 @@ class IntegrationGrid:
...
         Args:
             N (int): Total desired number of points in the grid (will take next lower root depending on dim)
             integration_domain (list or backend tensor): Domain to choose points in, e.g. [[-1,1],[0,1]]. It also determines the numerical backend (if it is a list, the backend is "torch").
+            grid_func (function): function for generating a grid of points over which to integrate (arguments: integration_domain, N, requires_grad, backend)
+            disable_integration_domain_check (bool): Disbaling integration domain checks (default False)
         """
         start = perf_counter()
-        self._check_inputs(N, integration_domain)
+        if not disable_integration_domain_check:
+            self._check_inputs(N, integration_domain)
```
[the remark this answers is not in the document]
ilan-gold: done! check it out!

torchquad/tests/integration_grid_test.py
```diff
@@ -1,14 +1,53 @@
...
 sys.path.append("../")
 
 from autoray import numpy as anp
 from autoray import to_backend_dtype
-
+import autoray as ar
 from integration.integration_grid import IntegrationGrid
+from integration.grid_integrator import GridIntegrator
+from integration.utils import _linspace_with_grads
+import tensorflow as tf
```
gomezzz, line 13: I think we meant to avoid direct imports as this means you can only run things in envs where you have the individual frameworks installed. As a compromise if necessary, maybe you can wrap it in a try / except call so the test at least doesn't fail if you don't have TF in this env?

gomezzz (approved):
only a minor comment, looks good :) :+1:

torchquad/tests/integration_grid_test.py
ilan-gold: thanks!

Issue #163, opened 2023-03-06
A lot of warnings in the current test CI on develop


gomezzz:
# Issue
## Problem Description

Currently tests throw a lot of warnings on `develop` branch

```
 =============================== warnings summary ===============================
torchquad/tests/boole_test.py: 4 warnings
torchquad/tests/gradient_test.py: 3 warnings
torchquad/tests/integrator_types_test.py: 4 warnings
torchquad/tests/monte_carlo_test.py: 4 warnings
torchquad/tests/simpson_test.py: 7 warnings
torchquad/tests/trapezoid_test.py: 4 warnings
  /home/runner/work/torchquad/torchquad/torchquad/tests/../integration/utils.py:255: UserWarning: DEPRECATION WARNING: In future versions of torchquad, an array-like object will be returned.
    warnings.warn(

torchquad/tests/boole_test.py::test_integrate_torch
  /home/runner/micromamba-root/envs/torchquad/lib/python3.10/site-packages/torch/functional.py:504: UserWarning: torch.meshgrid: in an upcoming release, it will be required to pass the indexing argument. (Triggered internally at /home/conda/feedstock_root/build_artifacts/pytorch-recipe_1675740247391/work/aten/src/ATen/native/TensorShape.cpp:3190.)
    return _VF.meshgrid(tensors, **kwargs)  # type: ignore[attr-defined]

torchquad/tests/monte_carlo_test.py::test_integrate_jax
  /home/runner/micromamba-root/envs/torchquad/lib/python3.10/site-packages/autoray/autoray.py:79: UserWarning: Explicitly requested dtype <class 'jax.numpy.float64'> requested in array is not available, and will be truncated to dtype float32. To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See https://github.com/google/jax#current-gotchas for more.
    return get_lib_fn(backend, fn)(*args, **kwargs)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================== 48 passed, 28 warnings in 84.06s (0:01:24) ==================
```

## Expected Behavior

Not or few warnings.

## What Needs to be Done

Fix warnings :v:

## How Can It Be Tested or Reproduced

Running tests with environment for all backends or on github.

gomezzz:
It has become even more, now > 100 warnings in CI. 😨 

```

=============================== warnings summary ===============================
torchquad/tests/boole_test.py: 6 warnings
[...]

PR #185, opened 2023-09-10
Hide PyTorch trace compilation warnings

FHof:
# Description

The test execution shows warnings about traces being potentially incorrect because the Python3 control flow is not completely recorded. 
This includes conditions on the shape of the integration domain tensor. 
Since the only arguments of the compiled integration function are the integrand and integration domain, 
and the dimensionality of this integration domain is constant, 
we can ignore the warnings.

After this change, the two `get_jit_compiled_integrate` functions hide PyTorch trace compilation warnings with `warnings.filterwarnings`.

Alternatively, it is possible to hide the warnings only at the code locations where they appear, but this would significantly increase the code complexity I think, see for example [this diff to hide a subset of the warnings](https://github.com/esa/torchquad/files/12568105/warnings_reduction.diff.txt).

## Resolved Issues

- [ ] It partly addresses #163

## How Has This Been Tested?

Execution of pytest before and after the changes


torchquad/integration/monte_carlo.py
```diff
@@ -196,6 +196,16 @@ def compiled_integrate(fn, integration_domain):
...
+                import warnings
+
+                # The PyTorch trace compilation warnings contain many false
+                # positives, so we hide all trace compiler warnings
+                def trace_without_warnings(*args, **kwargs):
+                    with warnings.catch_warnings():
+                        warnings.filterwarnings(
+                            "ignore", category=torch.jit.TracerWarning
+                        )
+                        return torch.jit.trace(*args, **kwargs)
```
gomezzz, line 208: I guess you are not making this function a dedicated file in the `utils` folder to avoid the torch import if possible, right?

gomezzz (approved):
Just a minor comment since ideally I would avoid having the same function defined in two places but probably unavoidable here due to imports?

FHof:
> Just a minor comment since ideally I would avoid having the same function defined in two places but probably unavoidable here due to imports?

I think the function can be moved without import problems. I've moved it now.

torchquad/integration/monte_carlo.py
FHof: I've moved it to the `utils.py` file now, which also contains helper functions.

PR #197, opened 2023-12-25
Fix JIT compilation with TensorFlow >= 2.14.0

### LINKED ISSUE #195: Regression in tests with TF
# Issue
## Problem Description

Boole test fails with

```
 =========================== short test summary info ============================
FAILED boole_test.py::test_integrate_tensorflow - tensorflow.python.framework.errors_impl.OperatorNotAllowedInGraphError: in user code:

    File "/home/runner/work/torchquad/torchquad/torchquad/tests/../integration/grid_integrator.py", line 131, in calculate_grid  *
        grid = IntegrationGrid(
    File "/home/runner/work/torchquad/torchquad/torchquad/tests/../integration/integration_grid.py", line 44, in __init__  **
        self._check_inputs(N, integration_domain, disable_integration_domain_check)
    File "/home/runner/work/torchquad/torchquad/torchquad/tests/../integration/integration_grid.py", line 114, in _check_inputs
        dim = _check_integration_domain(integration_domain)
    File "/home/runner/work/torchquad/torchquad/torchquad/tests/../integration/utils.py", line 209, in _check_integration_domain
        if boundaries_are_invalid:

OperatorNotAllowedInGraphError: Using a symbolic `tf.Tensor` as a Python `bool` is not allowed. You can attempt the following resolutions to the problem: If you are running in Graph mode, use Eager execution mode or decorate this function with @tf.function. If you are using AutoGraph, you can try decorating this function with @tf.function. If that does not work, then you may be using an unsupported feature or your source code may not be visible to AutoGraph. See https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/autograph/g3doc/reference/limitations.md#access-to-source-code for more information.
============= 1 failed, 51 passed, 75 warnings in 78.24s (0:01:18) =============
Error: Process completed with exit code 1.

```

## Expected Behavior

Test passing :)

## What Needs to be Done

Not quite sure, @FHof do you have an idea maybe? Decorating the entire is probably not possible given that requires tf install?

## How Can It Be Tested or Reproduced

Running boole test with TF.

FHof:
# Description

The previous check if code is currently being compiled no longer works with new TensorFlow versions because the `Tensor` type is now called `SymbolicTensor`.

This change adds a helper function to check if code is being compiled for JAX, TensorFlow or PyTorch. If tf.is_symbolic_tensor() is available, i.e. if the TensorFlow version is high enough, we use this function to check if code is being compiled.

To avoid inconsistencies between backends,
the check for integration domain values is disabled if code is being compiled with PyTorch even if the check works with PyTorch.

## Resolved Issues

Fixes #195

## How Has This Been Tested?

* There is a new test for the new function and the other tests are executed by the CI, too.
* I have manually tested `type(x.shape[0]).__name__ == "Tensor"` with PyTorch 2.0.1 but not PyTorch 1.10.0
* I have not tested if `_is_compiling` works for TensorFlow < 2.14.0; since it does the same as the previous condition in `_check_integration_domain`, I assume that it works.

### DIFF
--- a/torchquad/integration/utils.py
+++ b/torchquad/integration/utils.py
@@ -193,20 +193,11 @@ def _check_integration_domain(integration_domain):
             raise ValueError("integration_domain.shape[0] needs to be 1 or larger.")
         if num_bounds != 2:
             raise ValueError("integration_domain must have 2 values per boundary")
-        # Skip the values check if an integrator.integrate method is JIT
-        # compiled with JAX
-        if any(
-            nam in type(integration_domain).__name__ for nam in ["Jaxpr", "JVPTracer"]
-        ):
+        # The boundary values check does not work if the code is JIT compiled
+        # with JAX or TensorFlow.
+        if _is_compiling(integration_domain):
             return dim
-        boundaries_are_invalid = (
-            anp.min(integration_domain[:, 1] - integration_domain[:, 0]) < 0.0
-        )
-        # Skip the values check if an integrator.integrate method is
-        # compiled with tensorflow.function
-        if type(boundaries_are_invalid).__name__ == "Tensor":
-            return dim
-        if boundaries_are_invalid:
+        if anp.min(integration_domain[:, 1] - integration_domain[:, 0]) < 0.0:
             raise ValueError("integration_domain has invalid boundary values")
         return dim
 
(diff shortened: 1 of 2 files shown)


gomezzz (approved):
👍 👍 thanks so much :)

PR #199, opened 2024-04-25
Update `jax` config import

alberthli:
This PR changes the import statement for `jax.config`, which resolves the error
```
File "/usr/local/lib/python3.10/dist-packages/torchquad/utils/set_precision.py", line 58, in set_precision
    from jax.config import config
ImportError: cannot import name 'config' from 'jax.config' (/usr/local/lib/python3.10/dist-packages/jax/config.py)
```

### DIFF
--- a/torchquad/utils/set_precision.py
+++ b/torchquad/utils/set_precision.py
@@ -55,9 +55,9 @@ def set_precision(data_type="float32", backend="torch"):
         )
         torch.set_default_tensor_type(tensor_dtype)
     elif backend == "jax":
-        from jax.config import config
+        from jax import config
 
-        config.update("jax_enable_x64", data_type == "float64")
+        config.update("jax_enable_x64", True)
         logger.info(f"JAX data type set to {data_type}")
     elif backend in ["numpy", "tensorflow"]:
         os.environ[f"TORCHQUAD_DTYPE_{backend.upper()}"] = data_type


gomezzz:
Hi @alberthli , thanks for the PR! 

To understand a bit better, it is my understand jax [deprecated this import recently](https://github.com/google/jax/blob/36c471b6f52e50264f9b020995ecf176e7aecce4/jax/config.py#L18).

Did you check by any chance if this works for older jax version as well? Currently, we require `jax>=0.2.22` , so we may need to update the version 

(P.S. don't worry update the failing tests, the failures are unrelated to this PR, I think,  and a regression in the CI)

alberthli:
Hi @gomezzz, I haven't checked whether it works for later versions. [This commit](https://github.com/google/jax/commit/c90f1f0c960da87716d5c49e02efc4571e8a0781) from 6 months ago seems relevant if you want to control versioning, though.

alberthli:
Hi @gomezzz, my bandwidth is very limited in the next couple of weeks, so I probably won't be able to write this PR. I think separating the testing issues into a separate issue is a good idea, though perhaps that fix should be merged with this one together.

HGangloff:
Hi,
I would like to try torchquad with JAX for my project. Shouldn't we ask for a minimal JAX version of 0.4.25 following https://jax.readthedocs.io/en/latest/changelog.html#jax-0-4-25-feb-26-2024 ?
I can make the PR if needed

gomezzz:
Hi @HGangloff ,
ah yes, I was looking for that info unsuccessfully. I don't know jax so well :). 

Sounds good, please go ahead, thanks! In case you have an idea why the datatypes in the tests changed (which should have been set via the here modified `set_precision` function, I think) I would appreciate your insight!

gomezzz:
I will take care of this in #210 for release 0.4.1. Thanks again for pointing it out and proposing a fix!

PR #204, opened 2024-11-01
Feature/issue184 loguru debug

arkinn47:
# Description

Summary of changes

* In the __init__.py file, in the torchquad folder, line 49, which called the set_log_level function was deleted.  
* On lines 3 and 4, a commented line of explanation and the function "logger.disable("torchquad")" were added respectively.
* A test file logger_test.py was also added in the main folder, which runs a unit test testing if logger works.

## Resolved Issues

- [ ] fixes #Issue 184, where when torchquad is imported logger does not properly log ie. logger.info("..."), does not print anything.

## How Has This Been Tested?

- [ ] Test A: The test import torchquad and then attempts to write a log with logger.info and then checks if it was logged.

## Related Pull Requests

- #PR


gomezzz (changes requested):
Hi @arkinn47 !

Thanks for the PR, I think this could work! However, I think the problem is slightly more involved than this. Right now, we are using loguru also to display error messages to user via https://github.com/search?q=repo%3Aesa%2Ftorchquad%20logger.error&type=code and https://github.com/search?q=repo%3Aesa%2Ftorchquad+logger.warning&type=code .

- [ ] With logger.disable these would not be shown anymore, so I think we need to change those occurrences then to use something else? 
- [ ] Ideally we would set logger.disable only in release builds since we want logs for debugging. If we can't automate, we should at least mention it in contribute or similar? What do you think?

Thanks!

arkinn47:
If you remove the logger.disable("torchquad") function from line 4 then the error messages are displayed however, according to "Code Snippets and Recipes for Loguru" page, logger.disable("torchquad") should unconditionally be at the top of __init__.py to prevent the library's log handlers from also receiving user logs, which according to the website is undesirable. So either we can either simply remove  logger.disable("torchquad") or leave it in the release builds and inform users. You mentioned "change those occurrences," however I am not sure what we would change them to.

gomezzz:
Hi @arkinn47 ,
thanks for the update. I think it would make sense to replace the logger.error and logger.warning calls with ones using python's built-in `warnings` and for the errors just prints? That way users still get errors and warnings and we can just set disable as recommended? What do you think?

gomezzz:
closed in favor of #221

PR #205, opened 2024-11-05
Update set_precision.py

aurelio-amerio:
# Description
I believe `from jax.config import config` is a typo and it should be changed to `from jax import config`.

Summary of changes

* changed `from jax.config import config` -> `from jax import config`

## Resolved Issues

- solves an import issue with the jax backend


aurelio-amerio:
I hadn't realized there was a PR going on

PR #215, opened 2024-11-25
Fix Gaussian integrator on jax

### LINKED ISSUE #214: [jax] Gaussian and GaussLegendre throw errors
# Issue
## Problem Description

`Gaussian` and `GaussLegendre` integrators don't work with jax backend, they throw

```
TypeError: prod requires ndarray or scalar arguments, got <class 'list'> at position 0.
```

and this error comes from

https://github.com/esa/torchquad/blob/bf8ed5c1687e52d223cb8f55c0a99b2c8bbeaded/torchquad/integration/gaussian.py#L67-L69

## What Needs to be Done

I propose to convert the list returned by `anp.meshgrid` into an array with `anp.stack`. This fix works for me in jax, however we should first check that this doesn't cause problems in other backends.

```py
            return anp.prod(
                anp.stack(anp.meshgrid(*([weights] * dim), like=backend)), axis=0
            ).ravel()
```

## How Can It Be Tested or Reproduced

Using jax-0.4.35 and torchquad 0.4.0 run

```py
import jax
import jax.numpy as jnp
from torchquad import set_up_backend, MonteCarlo, Gaussian, GaussLegendre

set_up_backend(backend="jax")

@jax.jit
def some_function(x):
    return jnp.power(x[:, 0] - x[:, 1], 2)

g = Gaussian()
# It also fails with GaussLegendre
# g = GaussLegendre()

integral_value = g.integrate(
    lambda x: some_function(x),
    dim=2,
    N=10000,
    integration_domain=jnp.asarray([[-1.0, 1.0], [-1.0, 1.0]]),
)
```

javier-garcia-tilburg:
# Description

Gaussian and GaussLegendre integrators throw an error on jax if anp.prod is called on a list instead of an array

Summary of changes

* Before calling `anp.prod`, use `anp.stack` to convert list of arrays into an array.

## Resolved Issues

- [x] fixes #214

## How Has This Been Tested?

- [ ] Untested

**IMPORTANT**: This change could potentially affect the behavior of other backends.

### DIFF
--- a/torchquad/integration/gaussian.py
+++ b/torchquad/integration/gaussian.py
@@ -65,7 +65,7 @@ class Gaussian(GridIntegrator):
             ).ravel()
         else:
             return anp.prod(
-                anp.meshgrid(*([weights] * dim), like=backend), axis=0
+                anp.stack(anp.meshgrid(*([weights] * dim), like=backend)), axis=0
             ).ravel()
 
     def _roots(self, N, backend, requires_grad=False):


gomezzz:
Tested it locally also with other backends, seems to work :) Thanks!

gomezzz (approved):

PR #237, opened 2026-07-24
Add release-testing suite against latest backend versions

gomezzz:
## What & why

Adds a dedicated **`release_testing/`** suite — slower, end-to-end checks that run
against the **latest released** PyTorch / JAX / TensorFlow (not the pinned CI
environment), so a release cannot silently break against a new framework version.
The old docs claimed "tested with" specific backend versions that were never
actually exercised in a clean install; this closes that gap with real assertions.

Scenarios were drawn from a survey of real-world torchquad usage (particle/nuclear
physics, cosmology, Bayesian statistics, score-based ML, quantitative finance):

- **Core accuracy** — Trapezoid/Simpson/Boole/GaussLegendre on smooth 1D densities + a 3D separable integrand, every backend, float64, rule-appropriate tolerances.
- **Stochastic** — canonical 2D MonteCarlo (float32), 10D MC/VEGAS, seeded bit-for-bit determinism.
- **Autodiff** — gradient w.r.t. domain bounds (Leibniz) and w.r.t. a learned integrand parameter (the vebglm variational-Bayes pattern), torch + jax.
- **Complex** — complex-valued oscillatory Fourier kernels (characteristic functions), real/imag parts checked independently.
- **JIT** — `get_jit_compiled_integrate` vs eager parity (torch/JAX/TF).

All asserts use closed-form ground truth (`release_testing/_integrands.py`).

## Wiring

- `.github/workflows/release_testing.yml` — `workflow_dispatch` + on GitHub release, Python 3.10/3.11/3.12 matrix, installs latest backends.
- A step added to the release checklist (`.github/ISSUE_TEMPLATE/release.md`).

## Test plan

- [x] `pytest release_testing/` — **97 passed** locally against torch 2.5.1 / jax 0.4.35 / TF 2.18 / numpy 2.3 (Python 3.12)
- [x] `ruff check` + `ruff format --check` clean on the new files
- [x] `release_testing` workflow green (will trigger via workflow_dispatch after merge)

Part of the 0.6 release work.

### DIFF
--- a/.github/ISSUE_TEMPLATE/release.md
+++ b/.github/ISSUE_TEMPLATE/release.md
@@ -22,6 +22,7 @@ _to be written during release process_
 - [ ] Review the PR (if OK - merge, but DO NOT delete the branch)
 - [ ] On `Release` ,Minimize packages in requirements.txt and conda-forge submission. Update packages in pyproject.toml
 - [ ] Check unit tests -> Check all tests pass on CPU and [GPU (e.g. on colab)](https://colab.research.google.com/drive/1lFpdtY5zV7VpW88aazedA3n4khedHDQP?usp=sharing#scrollTo=IbU2vypPQ-Ej) and that there are tests for all important features
+- [ ] Run the extended **release-testing suite** against the latest framework versions: trigger the [Release testing GitHub Action](https://github.com/esa/torchquad/actions/workflows/release_testing.yml) on the release branch (or run `pytest release_testing/` locally in an env with the newest torch/JAX/TensorFlow) and confirm all checks pass. See [`release_testing/README.md`](https://github.com/esa/torchquad/blob/main/release_testing/README.md).
 - [ ] Check documentation -> Check presence of documentation for all features by locally building the docs on the release
 - [ ] Change version number in pyproject.toml and docs (under conf.py) and in `__init__.py`
 - [ ] In `__init__.py`, set `TORCHQUAD_DISABLE_LOGGING` to `True`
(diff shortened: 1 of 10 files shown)


gomezzz:
## `claude-utils` review — resolutions

Ran the structured review; all 5 findings addressed in `1bf303d`.

**1. (Important) Suite could go green while a backend fails to import** — every backend-parametrized test guards with `importorskip`, and the workflow ran plain `pytest`, so a broken latest-TF/JAX would silently skip and pass — the exact regression this suite exists to catch. **Fixed:** the workflow now installs `pytest-error-for-skips` and runs with `--error-for-skips` (matching the per-PR CI), so a missing/broken backend fails the job. Verified locally: 97 pass with the flag, i.e. no hidden skips in a full-backend env.

**2. (Important) "JIT vs eager parity" test didn't check parity** — with MonteCarlo the JIT and eager paths are independent RNG draws, so the test only checked each landed near the analytic value; a mis-scaled JIT result would pass. **Fixed:** rewrote it around `Simpson` (a deterministic rule, no RNG), so the compiled path must reproduce the eager result bit-for-bit (`rel_error < 1e-10`) *and* both hit ground truth. Real parity now.

**3. (Nit) `scipy.special.erf` for a scalar** — **Fixed:** use stdlib `math.erf`, dropped the scipy import from the ground-truth module.

**4. (Nit) Private `_setup_integration_domain` import** — **Fixed:** the JIT test now builds its backend-tensor domain via autoray's public API (`anp.array(..., like=backend)`), no internal coupling.

**5. (Nit) Bare `1e-12` threshold in `rel_error`** — **Fixed:** pulled into a named constant `_ZERO_DENOM_ATOL` with a rationale comment.

PR #238, opened 2026-07-24
Add packaging extras and modernize install docs (C5-C7)

gomezzz:
## Summary
- Adds `[project.optional-dependencies]` extras: `dev`, `docs`, and CPU-convenience backend extras (`torch`, `jax`, `tensorflow`, `all`). GPU builds stay out of scope for extras — PEP 508 cannot encode the per-index URLs each framework needs (C5).
- Rewrites the install docs (README + `install.rst`) to lead with pip/uv, drop the deprecated `pytorch` conda channel and the stale ROCm "investigating" line, and point GPU users at each framework's own guide (C6).
- Repairs `environment_all_backends.yml`: drops the dead `pytorch` channel, legacy `cudatoolkit`, and obsolete jax find-links; rebuilds on conda-forge (C7).

Part of the 0.6 packaging/CI modernization. Related to the roadmap items C5–C7.

## Test plan
- [x] `pyproject.toml` parses; all six extras resolve as valid PEP 508 requirement lists.
- [x] Both env YAMLs parse; channels reduced to `conda-forge` only.
- [x] CI (lint + all-backends tests) green on this branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,9 +93,9 @@ This is a brief guide for how to set up torchquad.
 
 ### Prerequisites
 
-We recommend using [conda](https://anaconda.org/conda-forge/torchquad), especially if you want to utilize the GPU.
-With PyTorch it will automatically set up CUDA and the cudatoolkit for you, for example.
-Note that torchquad also works on the CPU; however, it is optimized for GPU usage. torchquad's GPU support is tested only on NVIDIA cards with CUDA. We are investigating future support for AMD cards through [ROCm](https://pytorch.org/blog/pytorch-for-amd-rocm-platform-now-available-as-python-package/).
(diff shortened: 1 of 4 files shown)


gomezzz:
Claude: Thanks — addressed all three findings in abbd9dd (and propagated into the stacked #239):

- **Finding 1 (Important):** Corrected the misleading "(CPU)" labeling for `torch`. The pyproject block comment, README, and install.rst now state that `jax[cpu]`/`tensorflow` are CPU builds but default-index `torch` is the CUDA build on Linux (CPU on macOS/Windows), and point at the PyTorch CPU index for a guaranteed-CPU install.
- **Finding 2:** `all` is now a recursive extra — `torchquad[torch]` / `[jax]` / `[tensorflow]` — so each backend floor lives in exactly one place. Verified the wheel METADATA resolves the recursion correctly.
- **Finding 3:** Softened the `dev`-extra comment: only ruff/pydoclint/vulture/pytest floors match CI pins; the rest are reasonable minimums, not lockstep pins.

PR #239, opened 2026-07-24
CI overhaul: uv toolchain, matrix, and quality gates (C8-C12)

gomezzz:
## Summary
CI/toolchain overhaul (roadmap C8–C12). **Stacked on #238** — review/merge that first; this PR's base auto-retargets to `develop` once #238 merges.

- **uv toolchain (C8):** the test workflow now installs CPU wheels with `astral-sh/setup-uv`, the biggest CI speed/cost win. Adds `uv.lock` for reproducible dev envs.
- **pixi (C8b):** experimental `pixi.toml` with per-backend environments; not a CI dependency.
- **CI matrix (C9):** the `test` job runs all backends across Python 3.10/3.11/3.12 (`fail-fast: false`, `--error-for-skips`); the 3.12 leg carries the coverage gate + PR comment. Isolated JAX/TensorFlow jobs pin one backend. They keep a CPU torch baseline because several test modules `import torch` at collection time — full single-backend isolation waits on the 0.7 test-layout refactor.
- **Wheel smoke test (C10):** builds the wheel, installs it clean, runs the deployment self-test.
- **Docs gate (C11):** `sphinx-build -W`. Fixed the pre-existing warnings so it starts green — dropped the unsupported `display_version` theme option, switched `imgmath`→`mathjax` (no LaTeX on CI/RTD), and repaired a malformed `tutorial.rst` code block via `textwrap.dedent`.
- **Coverage floor (C12):** `--cov-fail-under=85` (measured ~89% locally).

## Test plan
- [x] `run_tests.yml` parses; `uv build` produces wheel + sdist.
- [x] `sphinx-build -W` builds clean (0 warnings) after the fixes.
- [x] Deployment self-test returns `True` from a clean wheel install (numpy only).
- [x] `textwrap.dedent` change verified to preserve column-0 runtime script.
- [x] Full matrix green on CI.


gomezzz:
Claude: Thanks — addressed all five findings in e4cf29d:

- **Finding 1 (Important):** The isolated jax/tf jobs now have an explicit `python -c "import <backend>"` step after install, so a pinned backend that fails to import fails the job instead of every test `importorskip`-ing to green — closing exactly the blind spot the job exists to catch.
- **Finding 2:** Fixed the comment — the `--cov-fail-under` floor runs on every matrix leg; only the PR coverage *comment* is 3.12-gated.
- **Finding 3:** Dropped the redundant severe-error ruff gate from the `test` matrix (it ran 3× and the parallel `lint` job's full `ruff check` already covers E9/F63/F7/F82).
- **Finding 4:** Added a comment in `pixi.toml` noting the core floors mirror `pyproject.toml` and must be bumped together (pyproject is the source of truth).
- **Finding 5:** Added a note in the wheel-smoke job flagging the `_deployment_test` → `deployment_test` rename coupling.

PR #240, opened 2026-07-24
Add Dependabot, CHANGELOG, and CONTRIBUTING (C13-C15)

gomezzz:
## Summary
Release-hygiene items (roadmap C13–C15). Independent of #238/#239.

- **Dependabot (C13):** `.github/dependabot.yml` keeps GitHub Actions on current majors via grouped weekly PRs. (OIDC trusted publishing was dropped — it needs a trusted-publisher configured on (Test)PyPI that doesn't exist yet, so the token-based deploy workflows are kept unchanged.)
- **CHANGELOG (C14):** `CHANGELOG.md` in Keep a Changelog format — 0.6 work under `Unreleased`, past releases backfilled.
- **CONTRIBUTING / branching (C15):** root `CONTRIBUTING.md` codifying the develop→main model (GitHub surfaces it at PR time) with a dev-setup section; README's duplicated contribution guide trimmed to a pointer. `main` is already a strict subset of `develop`, so no back-sync was needed.

## Maintainer note
`main` branch protection currently requires **0 status checks** — recommend marking `lint`/`test`/`dead_code` as required on `main` and `develop`.

## Test plan
- [x] YAML parses; net diff leaves the deploy workflows unchanged.
- [x] CI green.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  # Keep GitHub Actions pinned to current majors automatically. Grouped so all
+  # action bumps arrive in a single weekly PR instead of one per action.
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
--- a/.github/workflows/deploy_to_pypi.yml
+++ b/.github/workflows/deploy_to_pypi.yml
@@ -5,22 +5,22 @@ on: workflow_dispatch
 jobs:
   deploy:
     runs-on: ubuntu-latest
-    # The environment must match the trusted publisher configured on PyPI.
-    environment: pypi
-    permissions:
-      # OIDC token for PyPI trusted publishing — no stored API token needed.
-      id-token: write
-      # Declaring any permission drops the rest to none, so restore read access
-      # for checkout (don't rely on the repo staying public).
-      contents: read
 
     steps:
       - uses: actions/checkout@v4
-      - name: Set up uv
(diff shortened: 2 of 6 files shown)


gomezzz:
Claude: Thanks — addressed in 6ccdb55:

- **Finding 1:** Added `contents: read` to both deploy workflows so `checkout` doesn't silently rely on the repo being public (declaring any permission drops the rest to `none`).
- **Finding 2:** Added `skip-existing: true` to the Test PyPI publish so a re-triggered release for an already-published (immutable) version is a no-op rather than a red run.
- **Finding 3:** Added a comment noting `release/v1` is PyPA's intentionally auto-tracking ref that Dependabot can't bump — so the gap is deliberate, not an oversight.
- **Finding 4:** Kept `CHANGELOG.md`'s `Unreleased` as the 0.6 accumulator (standard Keep a Changelog usage). It's accurate on landing because the intended merge order is #238 → #239 → #240, so the extras/uv/matrix work is already on `develop` before this PR merges. Happy to reorder or hedge the wording if you'd rather merge this one first.

gomezzz:
Claude: Update per maintainer feedback (b69dcca): reverted the OIDC switch and kept the existing token-based twine deploy workflows — OIDC needs a trusted publisher configured on (Test)PyPI that isn't set up, so it would fail whereas the tokens work today. That makes review findings 1–3 (which hardened the OIDC workflows) moot; finding 4's changelog concern is unchanged. Also dropped the roadmap.md reference from the changelog and de-duplicated the README contribution guide against the new CONTRIBUTING.md. Net diff now leaves the deploy workflows untouched vs develop.

PR #241, opened 2026-07-24
refactor: dedupe grid_func and keep _deployment_test out of __all__

gomezzz:
## What

Two dead-code / public-API cleanups flagged in `REVIEW.md` for 0.6:

- **Dedupe `grid_func`.** `GridIntegrator._grid_func` returned a closure identical
  to the module-level `grid_func` in `integration_grid.py`. It now returns the
  shared function, so there is a single definition (Gaussian still overrides it).
- **Keep `_deployment_test` private, out of `__all__`.** The underscore is
  intentional — it's an internal post-release self-test that should not appear in
  the public API or user autocomplete. The real defect was an underscore name
  sitting in `__all__`; the fix is to drop it from `__all__` (not to rename it
  public). It stays reachable as `torchquad._deployment_test` for the wheel-smoke
  CI job and the README install check. Nothing user-facing depends on it, so no
  deprecation is needed.
- **Fix a circular import.** `deployment_test.py` imported from the top-level
  `torchquad` package while itself being imported during `__init__`; it now imports
  integrators and setup helpers from their submodules.

## Test plan

- [x] `ruff check`, `ruff format --check`, `pydoclint torchquad/`, `vulture ... --min-confidence 100` clean
- [x] `test_deployment.py` (incl. intent guard: reachable but not in `__all__`), `integration_grid_test.py` pass
- [x] `_deployment_test` accessible as `torchquad._deployment_test`, absent from `__all__`, no public `deployment_test`

Part of the 0.6 cleanup series (R6).

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -140,7 +140,7 @@ jobs:
           uv venv --python 3.12
           source .venv/bin/activate
           uv pip install dist/*.whl
-          python -c "import torchquad; assert torchquad.deployment_test()"
+          python -c "import torchquad; assert torchquad._deployment_test()"
 
   # Documentation build gate: warnings are errors so doc rot fails the PR
   # instead of silently shipping to Read the Docs.
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -140,9 +140,7 @@ jobs:
           uv venv --python 3.12
           source .venv/bin/activate
           uv pip install dist/*.whl
-          # NOTE: _deployment_test is slated to be renamed to deployment_test
-          # (drop the underscore); update this line in that rename PR.
-          python -c "import torchquad; assert torchquad._deployment_test()"
+          python -c "import torchquad; assert torchquad.deployment_test()"
 
   # Documentation build gate: warnings are errors so doc rot fails the PR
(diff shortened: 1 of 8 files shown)


gomezzz:
Claude: Thanks — both nits addressed in 5c8bead:

1. The deprecation regression test now exercises `torchquad._deployment_test()` (the top-level access path users rely on), so a break in the `__init__` re-export is caught, not just the submodule symbol.
2. Folded the duplicated alias comment into the single import comment and dropped the standalone block after `__all__`.

gomezzz:
Claude: Changed approach on the deployment self-test per maintainer feedback (69b22f6):

Instead of renaming `_deployment_test` → public `deployment_test`, it now **stays private and is simply removed from `__all__`**. The leading underscore was deliberate — it keeps this internal post-release self-test out of the public API and out of user autocomplete (a non-underscore name in `__all__` does surface in IPython/IDE completion). The actual REVIEW.md defect was the contradiction of an underscore name living in `__all__`, which dropping it from `__all__` resolves. Nothing user-facing depends on it (only the wheel-smoke CI job + tests), so the deprecation alias was dropped entirely.

`torchquad._deployment_test()` remains reachable for the CI self-test and the README install check. The `grid_func` dedup and circular-import fix are unchanged. Title/description updated.

PR #242, opened 2026-07-24
docs: fix stale badges and Built With tooling line

gomezzz:
## What

Truth-pass tail for the 0.6 docs work:

- The docs index carried a **dead Actions badge** using the removed
  `img.shields.io/github/workflow/status/...` endpoint. Replaced with the current
  `run_tests.yml` badge (same one the README uses).
- Every docs badge `:target:` pointed at the shields.io **image URL**, so clicking
  a badge showed raw SVG. Repointed each to its real destination (Actions runs,
  PyPI project, conda-forge, GitHub commits/issues/pulls/contributors/LICENSE).
- README "Built With" no longer implies conda is the only environment tool; it now
  mentions uv/conda, consistent with the pip/uv-lead install section.

Verified torchquad is in fact published on conda-forge (0.5.0), so the conda
badges are kept.

## Test plan

- [x] `sphinx-build -W -b html docs/source` builds clean (docs CI gate)

Part of the 0.6 cleanup series (R7).

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ The software is free to use and is designed for the machine learning community a
 This project is built with the following packages:
 
 * [autoray](https://github.com/jcmgray/autoray), which means the implemented quadrature supports [NumPy](https://numpy.org/) and can be used for machine learning with modules such as [PyTorch](https://pytorch.org/), [JAX](https://github.com/google/jax/) and [Tensorflow](https://www.tensorflow.org/), where it is fully differentiable
-* [conda](https://docs.conda.io/en/latest/), which will take care of all requirements for you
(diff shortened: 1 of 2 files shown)


gomezzz:
Claude: Thanks!

- Finding 1: fixed in faab1e7 — the last-commit badge now targets `/commits/main` explicitly, consistent with the other GitHub targets.
- Finding 2: left deliberate. The default-branch `badge.svg` (no `?branch=`) matches the README Tests badge exactly, and the default branch (`main`) is what the docs deploy from, so `main`'s status is the right thing to show here. Kept without `?branch=` to stay in sync with the README.

Issue #180, opened 2023-06-04
torch.tensor containing integers as integration domain returns zero with non-compiled integrator


julian-urban:
Hi,

first of all thanks a lot for this great package, I've been using it heavily over the past few weeks and I'm very happy about its performance.

The issue: today I started using the functionality for calculating derivatives with respect to domain boundaries. I had previously employed jit-compiled integrators which require torch tensors as arguments for the integration domains. Compilation currently seems to be incompatible with computing the gradients, so I switched back to the non-compiled integrator, leaving everything else untouched. I then noticed that the integration consistently returns zero unless I pass the integration domain as a list instead of a torch tensor.

Example code to reproduce:

```
integrator = Simpson()
print(integrator.integrate(
    lambda x: x,
    dim=1,
    N=101,
    integration_domain=[[0,1]],
    backend='torch'))
print(integrator.integrate(
    lambda x: x,
    dim=1,
    N=101,
    integration_domain=torch.tensor([[0,1]]),
    backend='torch'))
jit_integrator = integrator.get_jit_compiled_integrate(
    dim=1,
    N=101,
    backend='torch')
print(jit_integrator(
    lambda x: x,
    torch.tensor([[0,1]])))
```

This returns the values 0.5, 0., 0.5, but they should of course all be the same. This is unexpected behavior, it should at least throw an error / a warning to inform the user about the incorrect type. Although IMHO it should ideally allow lists, numpy arrays, and torch tensors, for the integration domain argument. It's a minor issue but may lead to confusion (as in my case), so it might be a good idea to catch this somehow.

Cheers

Edit: I found that if I use ```torch.tensor([[0.,1.]])``` instead of ```torch.tensor([[0,1]])```, it does work as expected. So the problem is not the type of the object, but that it contains integers instead of floats. I still believe this should be fixed or a warning issued. I'm not sure why the jit-compiled version is not affected by this bug.

gomezzz:
Should be fixed with #181 which will be part of Release 0.4.0 ! :) 

Thanks for pointing it out!

PR #243, opened 2026-07-24
test: add cross-backend agreement and integer-domain regression tests

gomezzz:
## What

Two correctness properties from `REVIEW.md` section 1 that were untested:

- **Cross-backend agreement.** `test_cross_backend_agreement` runs Trapezoid,
  Simpson, Boole and GaussLegendre over a 1D and a 2D integrand on every
  installed backend in fp64 and asserts they agree to ~round-off. Both integrands
  are linear per dimension (exact for these rules), so the test also anchors to
  the analytic value — a shared-but-wrong result cannot pass. The 2D case
  exercises `meshgrid` (issue #214).
- **Integer-domain regression (#180).** `test_integer_domain_*` passes an
  integer-dtype `integration_domain` to each grid integrator and asserts a
  correct, non-zero result, guarding the int→float promotion in `IntegrationGrid`.

Stochastic integrators (MonteCarlo, VEGAS) are intentionally excluded from the
agreement test — they draw from per-backend RNGs and can't match bit-for-bit.

## Test plan

- [x] `pytest cross_backend_test.py` — 5 passed across numpy/torch/jax/tensorflow
- [x] `ruff check` / `ruff format --check` clean

Part of the 0.6 cleanup series (R10).

### DIFF
--- a/tests/cross_backend_test.py
+++ b/tests/cross_backend_test.py
@@ -10,6 +10,8 @@ between backends (see REVIEW.md section 1):
   (issue #180); it has to be promoted to float and produce the correct result.
 """
 
+from importlib.util import find_spec
+
 import numpy as np
 import pytest
 import autoray as ar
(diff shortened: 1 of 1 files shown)


gomezzz:
Claude: Thanks — all three addressed in 2f50a5c:

1. **Backend order (Important):** `_ALL_BACKENDS` is now `["numpy", "torch", "tensorflow", "jax"]` with a comment referencing the anti-OOM rationale from `utils_integration_test.py`. The `test_integer_domain_*` definitions are reordered to match (JAX last).
2. **Availability probe:** replaced `__import__` with `importlib.util.find_spec`, derived from the single ordered `_ALL_BACKENDS` list (one source of truth, no side-effecting full import).
3. **Global state:** `test_cross_backend_agreement` now restores the default numpy/float32 global state in a `finally` block, so later tests relying on the default are unaffected.

Re-ran: 5 passed across numpy/torch/jax/tensorflow; ruff clean.

PR #244, opened 2026-07-24
feat: add Sobol quasi-Monte Carlo sampler

gomezzz:
## What

First-class quasi-Monte Carlo via a scrambled **Sobol** sampler that plugs into
`MonteCarlo.integrate` through the existing `rng=` slot — no changes to
`MonteCarlo` itself:

```python
from torchquad import MonteCarlo, Sobol
mc = MonteCarlo()
mc.integrate(fn, dim=3, N=2**13, integration_domain=[[0,1]]*3,
             rng=Sobol(backend="torch", seed=0))
```

For smooth integrands the error shrinks close to `O(1/N)` instead of plain Monte
Carlo's `O(1/sqrt(N))`. In the tests, at N=2¹³ the Sobol error is ~3e-10 vs
~1e-4…1e-3 for pseudo-random MC across all four backends.

- **torch**: `torch.quasirandom.SobolEngine` (native; points moved onto the
  current default device so GPU domains line up, matching `RNG`'s `torch.rand`).
- **numpy / jax / tensorflow**: `scipy.stats.qmc.Sobol` (SciPy is already a hard
  dependency) converted to the backend.
- Sample points are constants, so autodiff through the integrand and the domain
  is preserved exactly as for plain MC.
- The class mirrors `RNG`'s per-backend shape (the established exception to the
  backend-agnostic rule for RNG-like samplers) and is exposed as `torchquad.Sobol`.

## Notes / scope

- Targets the eager `integrate` path; the JIT-compiled path builds its own RNG
  and is out of scope here.
- Per-backend Sobol implementations scramble differently, so results are
  reproducible for a fixed seed *within* a backend but not bit-for-bit across
  backends (documented on the class).

## Test plan

- [x] `sobol_test.py`, all backends: analytic accuracy, Sobol < plain-MC error at
  equal N (deterministic, fixed seed), seed reproducibility, torch gradient flow — 13 passed
- [x] `monte_carlo_test.py` still green (19 passed together)
- [x] `ruff` / `pydoclint` / `vulture` clean; `sphinx-build -W` builds (Sobol auto-documented via `__all__`)

Addresses #140 (MonteCarlo path; VEGAS + JIT support to follow). Motivated by #217.

Roadmap F1.

PR #245, opened 2026-07-24
feat: add opt-in VEGAS error reporting (return_error)

gomezzz:
## What

VEGAS already computes a standard deviation, chi-squared and goodness-of-fit
internally, then threw them away and returned only the bare integral. This adds
an opt-in `return_error=True` that surfaces them:

```python
from torchquad import VEGAS
result = VEGAS().integrate(fn, dim=2, N=40000, integration_domain=dom, return_error=True)
print(result)        # VEGASResult(4.00144 +/- 0.0014, chi2/dof = 1, Q = 0.41)
result.integral, result.sdev, result.chi2, result.dof, result.Q
```

- New `VEGASResult` dataclass (`integral`, `sdev`, `chi2`, `dof`, `Q`,
  `nr_of_fevals`) with a compact `value +/- sdev (chi2/dof, Q)` repr; `Q` is the
  chi-squared goodness-of-fit p-value via SciPy's regularized upper incomplete
  gamma. Exposed as `torchquad.VEGASResult`.
- The tensor fields keep the integration backend. Only `integral` is on the
  gradient path; `sdev`/`chi2` derive from VEGAS's detached variances and are
  not differentiable.
- `return_error` defaults to `False` → the existing bare-value return is
  unchanged (no breaking change).

## Test plan

- [x] `vegas_result_test.py` (numpy + torch): result fields well-formed and
  self-consistent, error estimate brackets the true error, backward-compatible
  default return; numpy bit-for-bit reproducibility (torch bit-reproducibility
  isn't asserted — GPU reductions aren't deterministic)
- [x] existing `vegas_test.py` still green
- [x] `ruff` / `pydoclint` / `vulture` clean; `sphinx-build -W` builds
  (`VEGASResult` auto-documented via `__all__`, per-field attribute docstrings)

Roadmap F2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ The 0.6 line is a modernization and credibility release: modern tooling, honest
 packaging, and closing long-open fixed issues.
 
 ### Added
+- `VEGAS.integrate(..., return_error=True)` returns a `VEGASResult` bundling the
+  integral with its error estimate (standard deviation, chi-squared, degrees of
+  freedom and goodness-of-fit Q) instead of discarding them.
 - Optional-dependency extras: `dev`, `docs`, and CPU-convenience backend extras
   `torch`, `jax`, `tensorflow`, `all`.
 - `release_testing/` suite — slower end-to-end checks run against the latest
--- a/tests/vegas_result_test.py
+++ b/tests/vegas_result_test.py
@@ -52,9 +52,14 @@ def _vegas_result_fields_test(backend, dtype_name=None):
         f"Error estimate {sdev} does not bracket the true error {abs(integral - _EXPECTED)}"
     )
 
-    # Degrees of freedom and goodness-of-fit are well-formed.
+    # Degrees of freedom and goodness-of-fit are well-formed, and Q matches an
+    # independent evaluation of the chi-squared survival function (guards against
+    # a swapped-argument or wrong-formula regression, not just a plausible range).
+    from scipy.special import gammaincc
+
     assert result.dof >= 1
     assert result.Q is not None and 0.0 <= result.Q <= 1.0
+    assert abs(result.Q - float(gammaincc(result.dof / 2.0, chi2 / 2.0))) < 1e-12
     assert result.nr_of_fevals > 0
     # __repr__ must not raise.
     assert "VEGASResult" in repr(result)
--- a/tests/vegas_result_test.py
+++ b/tests/vegas_result_test.py
@@ -1,0 +1,123 @@
+"""Tests for VEGAS error reporting via ``return_error=True``.
+
+VEGAS already computes an error estimate, chi-squared and goodness-of-fit
+internally; ``return_error=True`` surfaces them as a :class:`VEGASResult` instead
+of discarding them. VEGAS supports only the numpy and torch backends.
+"""
+
+import numpy as np
+from autoray import numpy as anp
+
+from torchquad.integration.vegas import VEGAS
+from torchquad.integration.vegas_result import VEGASResult
+from helper_functions import setup_test_for_backend
+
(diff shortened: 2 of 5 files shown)


gomezzz:
Claude: Thanks — all three addressed in 467d25f:

1. **Differentiability (Important):** narrowed the claim in the `VEGASResult` docstring and the PR description — only `integral` is on the gradient path; `sdev`/`chi2` come from VEGAS's detached per-iteration variances and are not differentiable.
2. **_get_chisq zero-variance:** added the `and sig2 != 0.0` guard (matching `_get_error`/`_get_result`) so the now-user-facing `chi2`/`Q` can't become inf/nan on a zero-variance iteration.
3. **Q oracle:** the fields test now asserts `result.Q == gammaincc(dof/2, chi2/2)` from the returned `chi2`/`dof`, so a swapped-argument or wrong-formula regression fails instead of passing a range check.

Re-ran vegas_test.py + vegas_result_test.py: 8 passed.

gomezzz:
Claude: ### Accuracy comparison: Monte Carlo vs Sobol QMC vs VEGAS

Relative error against the analytic value at an equal sample budget of **N = 2¹⁴ = 16384**, numpy backend, float64, fixed `seed=0`. Every integrand is separable so the exact integral is known in closed form.

| Integrand | dim | Monte Carlo | Sobol QMC | VEGAS |
|---|---|---|---|---|
| smooth: ∏ cos(πx/2) | 2 | 3.84e-03 | **4.53e-09** | 4.33e-04 |
| smooth: ∏ e^(-x) | 5 | 4.84e-04 | **8.55e-06** | 8.15e-05 |
| sharply peaked: ∏ e^(-((x-0.5)/0.1)²) | 3 | 2.48e-02 | 4.17e-03 | **2.46e-03** |
| smooth: ∏ e^(-x) | 8 | 1.34e-03 | **1.11e-04** | 3.46e-04 |

**Takeaways**
- On **smooth** integrands Sobol QMC is the clear winner — up to ~6 orders of magnitude better than plain MC (and consistently ahead of VEGAS), matching the expected O(1/N) vs O(1/√N) convergence.
- On the **sharply peaked** integrand, VEGAS's adaptive stratification pays off and it edges out Sobol; Sobol still beats plain MC ~6×, but its low-discrepancy advantage shrinks when the mass is concentrated in a small region.
- Plain Monte Carlo is the weakest in every case.

Caveat: VEGAS's `N` is an approximate *max* function-evaluation budget spread across warmup + refinement iterations, so its effective sample count isn't identical to MC/Sobol — this is a same-budget comparison, not an identical-sample-count one.

gomezzz:
Claude: Added an error-**calibration** test against the analytic collection (c3b18c2) — thanks, the earlier tests only exercised one integrand.

`vegas_result_test.py::_vegas_error_calibration_test` now runs **every real, scalar-integrand** function from `integration_test_functions` (1-D and 3-D; VEGAS doesn't support complex or multi-integrand functions, so those are skipped) with `return_error=True`, and for each asserts:
- VEGAS converged to the closed-form value (worst observed rel error ~0.8%);
- its self-reported `sdev` **brackets the true error**: `|integral − exact| < 8·sdev`.

Measured `true_error/sdev` across the collection: max ~3.4, mean ~1 on both numpy and torch — i.e. the estimate is a genuine 1σ-scale error bar, not just a plausible positive number. The 8× bound leaves headroom for VEGAS's stochastic (and GPU-nondeterministic) spread.

Runs on numpy and torch. Related tracked guidance is in the REVIEW.md update on #244.

gomezzz:
Claude: Merged `develop` in to resolve conflicts after #246 landed — VEGAS `integrate()` now carries both `return_error` (this PR) and `args` (from #246), verified together (vegas + result + args tests, 22 passed). CI green. Includes the error-calibration test against the analytic collection.

PR #188, opened 2023-10-03
grid_integrator: Allow passing custom arguments to integrand function.

corepattern:
This allows avoiding the invocation of a lambda wrapper function. 
The commit just exposes an already existing functionality.

Also related to #187

Issue #186, opened 2023-09-28
Example/documentation for parametric domain of integration?


mawright:
Hi,

I found issue https://github.com/esa/torchquad/issues/170 asking about the case where both the integrand and the domain of integration are functions of parameters, and it looks like @ilan-gold implemented this feature in pull request https://github.com/esa/torchquad/pull/173. I'd like to use this feature and wanted to ask if there are any examples or documentation showing how to use it.

Thanks!

gomezzz:
Hey @mawright ! 

Thanks for your interest! This should be documented [here in the readthedocs](https://torchquad.readthedocs.io/en/main/tutorial.html#multidimensional-vectorized-integrands), I think? :)

In case it is what you are looking for, feel free to suggest any potential changes that can help make it easier to find. 🙏

ilan-gold:
Hi @mawright - to clarify, what I did is make it so that you **could** implement this by tweaking a few things in the API.  There's no feature at the moment for it.  You would have to override a few functions to handle however you want to handle the new domain.

ilan-gold:
I can perhaps write something up explaining which methods to override - how does that sound @gomezzz ?  Perhaps an example if it's not too crazy?

mawright:
Thanks! Looking forward to it!

mawright:
I wrote something that works for my problem. It's here: https://github.com/mawright/torchstable/blob/main/torchstable/integrator.py.

ilan-gold:
@mawright I am sorry for not doing this yet.  I'm finishing my thesis at the moment so have been a bit busy.  You got it right, though, it seems.  My code looks quite similar

mawright:
No problem. One thing I was thinking is that it might be easier to work with batch integration if there was an additional batch dimension. The hacked-together thing I wrote breaks a lot of the code's assumptions as to the meaning of the tensor dimensions, e.g. that the polynomial roots and the number of rows of `integration_domain` have the same dimension, the splitting up of the `N` gridpoints over a multidimensional space, etc. Not sure how much work that woul be for you though.

gomezzz:
[...]

PR #246, opened 2026-07-24
feat: forward integrand args via integrate(..., args=...)

### LINKED ISSUE #187: Integrate function with parameters
Hi, I would like to know whether there is functionality for integrating functions with parameters. For example, if one has a function

```
def func(alpha, x):
    return x**2 * alpha
```

how can one apply `Trapezoid` rule to integrate over `x`, for example? In the docs there is information about additional arguments of functions in `BaseIntegrator` class, but it seems like it is always set to `None` when applied to concrete methods. 

If there is no such option, one may add something like `args=(alpha,)` to `Trapezoid.integrate()` or any other method. 

Thanks in advance!

gomezzz:
## What

Lets a parametric integrand be integrated directly, without a lambda wrapper:

```python
def fn(x, alpha, beta):
    return alpha * x[:, 0] + beta

Trapezoid().integrate(fn, dim=1, N=101, integration_domain=[[0, 1]], args=(alpha, beta))
```

`BaseIntegrator.evaluate_integrand` already accepted `args` and called
`fn(points, *args)`, but no public `integrate()` exposed it. This threads an
`args` parameter through every integrator — Trapezoid, Simpson, Boole,
GaussLegendre, MonteCarlo and VEGAS.

- `args` defaults to `None` (`fn(points)`), so all existing calls are unchanged.
- Scope is the eager `integrate()` path; the JIT-compiled path is untouched.

This re-implements the stale PR #188 by Dan Barzilay (credited as co-author) and
closes #187.

## Test plan

- [x] `args_test.py`: every integrator forwards `args` and integrates the
  parametric integrand correctly across backends (grid rules exact; MC/VEGAS
  within tolerance); `args=None` matches a pre-bound lambda — 12 passed
- [x] full integrator sweep (trapezoid/simpson/boole/gauss/monte_carlo/vegas/
  integrator_types) still green — 46 passed
- [x] `ruff` / `pydoclint` / `vulture` clean

Closes #187. Roadmap F3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ The 0.6 line is a modernization and credibility release: modern tooling, honest
 packaging, and closing long-open fixed issues.
 
 ### Added
+- `args` argument on every integrator's `integrate()` — extra parameters are
+  forwarded to the integrand as `fn(points, *args)`, so parametric integrands no
+  longer need a lambda wrapper (#187, #188).
 - Optional-dependency extras: `dev`, `docs`, and CPU-convenience backend extras
   `torch`, `jax`, `tensorflow`, `all`.
 - `release_testing/` suite — slower end-to-end checks run against the latest
--- a/tests/args_test.py
+++ b/tests/args_test.py
@@ -1,0 +1,113 @@
+"""Tests for passing extra integrand arguments via ``args`` (issues #187, #188).
+
+Every integrator forwards ``args`` to the integrand as ``fn(points, *args)``, so a
+parametric integrand can be integrated without wrapping it in a lambda.
+"""
+
+import numpy as np
+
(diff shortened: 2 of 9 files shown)


gomezzz:
Claude: Thanks!

- **Finding 2:** added a comment in `vegas.py` explaining that VEGAS bakes the integrand into a closure (and evaluates via `_eval`) rather than routing `args` through `evaluate_integrand`, so the `None -> ()` normalization is intentionally local — with a note to keep the two in sync (1d97cac).
- **Finding 1 (tutorial):** deferring to the F4 docs bundle, whose 'passing parameters' section (issue #186) is exactly the discoverability fix #187 asked for. I'll add the `args=(alpha, beta)` parametric example there alongside the QMC and VEGAS-error snippets so all the new-feature docs land together.

Issue #217, opened 2025-05-10
Quasi-Monte Carlo Need


Yuer-1218:
I'm sorry to bother, But I have a question about maybe there can add easy QMC, It helps with faster convergence, sorry  for my poor code ability that I have tried and failed, I don't know maybe it could be hard, Thanks a lot

PR #249, opened 2026-07-30
feat(qmc): add Lattice quasi-Monte Carlo sampler

### LINKED ISSUE #140: Low Discrepancy Sequences for MonteCarlo and VEGAS
# Feature

Low discrepancy sequences could be used as an optional replacement for random number generation for MonteCarlo and VEGAS.
With certain integrands, this could lead to a higher average accuracy. 

## What Needs to Be Done

* Add a class similar to the RNG class to generate numbers with a low discrepancy sequence instead of a PRNG. An instance of it can be passed as `rng` argument to VEGAS and MonteCarlo.
  This class could use, for example, PyTorch's and TensorFlow's sobol sequences: https://www.tensorflow.org/api_docs/python/tf/math/sobol_sample, https://pytorch.org/docs/stable/generated/torch.quasirandom.SobolEngine.html
* Add a function to the number generator classes which samples points and use it for MonteCarlo and VEGAS instead of `uniform`. In comparison to `uniform`, the output of this function always corresponds to points in a space, where conceptually a distance function is defined.
* Change `MonteCarlo.get_jit_compiled_integrate` so that it works with the number generator class for low discrepancy sequences.

## How Can It Be Tested

It can be tested with additional tests in the torchquad/tests folder.

Samuel-Vangu:
feat(qmc): add Lattice quasi-Monte Carlo sampler

# Description
Summary of changes
* Adds a randomly shifted rank-1 lattice quasi-Monte Carlo sampler, shaped like `RNG` (and mirroring the existing `Sobol` sampler), that plugs into `MonteCarlo.integrate` via the existing `rng=` slot.
* Points are generated from a component-by-component (CBC) generating vector (`lattice-33002-1024-1048576.9125`, from Frances Y. Kuo's *Lattice Rule Generating Vectors* collection, order-3 weights, dimensions up to 9125, sample sizes from 1024 to 1048576) as `x_i = frac(i * z / N)`.
* A random shift modulo one is applied on every call to `uniform` to keep the QMC estimator unbiased. With a fixed `seed`, the shift is reproducible across calls; without one, each call draws its own independent shift.
* torch backend only for now — the modular arithmetic and shift use `torch.Generator`/tensor ops directly; there is no SciPy/other-backend path like Sobol's, since `scipy.stats.qmc` has no rank-1 lattice rule.
* Sample points are constants, so autodiff through the integrand and domain is preserved exactly as with plain Monte Carlo (and with Sobol).
* Exposed as `torchquad.Lattice`.

Addresses [#140](https://github.com/esa/torchquad/issues/140) (MonteCarlo path; VEGAS and JIT support to follow). Motivated by the QMC request in [#217](https://github.com/esa/torchquad/issues/217).

## Resolved Issues
- [x] fixes #140

## How Has This Been Tested?
- [x] Analytic accuracy on a smooth integrand
- [x] Lattice beating plain Monte Carlo at equal N
- [x] Seed reproducibility (fixed seed ⇒ identical result)
- [x] Torch gradient-flow check (autodiff survives through the integral)
- [x] Closed-form check: unshifted points match `frac(i*z/N)` computed independently from the on-disk generating vector
- [x] Shift behaviour: same seed reproduces the shift across calls; no seed draws an independent shift each call; `shift=False` ignores the seed entirely
- [x] Input validation: N/dim range checks, power-of-two warning, whole-number/bool type checks

## Related Pull Requests
- #244


Samuel-Vangu:
Hi @gomezzz, could you please trigger the CI for this PR when you have time? Thank you!

Samuel-Vangu:
@gomezzz, pushed a few fixes (ruff formatting + moving `import torch` local to
Lattice.uniform) — could you trigger the CI again when you get a chance? Thanks!

Samuel-Vangu:
Hi @gomezzz, sorry for the back-and-forth — pushed another fix, could you trigger CI once more when you get a chance? Thanks for your patience!

Samuel-Vangu:
Hi @gomezzz, pushed pydoclint fixes ,  could you approve/trigger CI  when you get a chance? Thanks!

torchquad/integration/lattice-33002-1024-1048576.9125
gomezzz, line 1: This is a fairly large file blowing up the diff, probably should use git-lfs if needed? It also currently wouldn't ship because it is not packaged in pyproject.toml

torchquad/integration/qmc.py
```diff
@@ -91,3 +91,209 @@ def uniform(self, size, dtype):
...
+    Returns
+    -------
+    z : np.ndarray
+        NumPy array [z1, z2, ..., zd] with dtype uint64.
+    """
+    path = DATA_DIR / filename
+    if d <= 0:
+        raise ValueError("The dimension d must be strictly positive.")
+
+    data = np.loadtxt(path, dtype=np.uint64)
```
gomezzz: The full 146 KB / 9125-row file is re-read and re-parsed on every `uniform()` call, even for
`d=1`. Small against one 8192-point integration, but this is a
per-call cost. Would suggest to wrap the read in `functools.lru_cache` (cache the full array once, slice `[:dim]` per call).

torchquad/integration/qmc.py
```diff
@@ -91,3 +91,209 @@ def uniform(self, size, dtype):
...
+        Returns:
+            torch tensor: ``[number_of_points, dim]`` lattice points in
+            ``[0, 1)``.
+
+        Raises:
+            ValueError: If the number of points is not between 1024 and
+                1048576, or if the dimension is not between 1 and 9125, or if
+                either is not a whole number.
+        """
+        import torch
```
gomezzz: `Lattice` has no backend argument and no backend check, so passing it to a numpy/jax/tensorflow
domain fails deep inside torch rather than at the API boundary:

    TypeError: to() received an invalid combination of arguments - got
    (numpy.dtypes.Float64DType), but expected one of: ...

For comparison, `Sobol` takes
`backend` in its constructor. Would suggest to mirror it, raise a clear `ValueError` for anything but `"torch"`,
document the `Raises:` entry, and add a test for the error.

torchquad/integration/qmc.py
```diff
@@ -91,3 +91,209 @@ def uniform(self, size, dtype):
...
+
+        number_of_points = _check_positive_whole_number(size[0], "The number of points")
+        dim = _check_positive_whole_number(size[1], "The dimension")
+
+        if not 1024 <= number_of_points <= 1048576:
```
gomezzz: 1024 / 1048576 / 9125 are hard-coded here, restated in the class docstring, restated in the
test file, and encoded in the data filename, four places to keep in sync if the vector is ever
swapped. Would suggest to define module constants (`_MIN_POINTS`, `_MAX_POINTS`, `_LATTICE_FILE`) and derive the
dimension bound from `data.shape[0]` at load time.

torchquad/integration/qmc.py
```diff
@@ -91,3 +91,209 @@ def uniform(self, size, dtype):
...
+        if not 1 <= dim <= 9125:
+            raise ValueError(f"The dimension must be between 1 and 9125, but got {dim}.")
+
+        if number_of_points & (number_of_points - 1) != 0:
+            warnings.warn(
+                "The simple lattice construction used here requires the number "
+                f"of points to be a power of 2, but {number_of_points} was "
+                "requested.",
+                stacklevel=2,
+            )
```
gomezzz, line 327: Consider raising instead of warning. The docstring says the construction "is valid when the
total number of points is an exact power of two", a non-power-of-two N isn't a slightly-worse
variant, it's outside the rule's validity.

torchquad/integration/qmc.py
```diff
@@ -91,3 +91,209 @@ def uniform(self, size, dtype):
...
+    d : int
+        Desired dimension.
+    filename : str
+        Path to the generating vectors file.
+
+    Returns
+    -------
+    z : np.ndarray
+        NumPy array [z1, z2, ..., zd] with dtype uint64.
+    """
```
gomezzz: docstring style here does not match ,please use Google style :)

torchquad/integration/qmc.py
```diff
@@ -91,3 +91,209 @@ def uniform(self, size, dtype):
         sampler = qmc.Sobol(d=dim, scramble=self._scramble, seed=self._seed)
         points = sampler.random(number_of_points)
         return anp.array(points, dtype=dtype, like=self._backend)
+
+
+import numpy as np
+from pathlib import Path
```
gomezzz: please keep imports at the top

torchquad/integration/qmc.py
```diff
@@ -91,3 +91,209 @@ def uniform(self, size, dtype):
...
+        raise ValueError(
+            f"Requested dimension d={d}, but the file contains only {max_d} components."
+        )
+
+    z = data[:d, 1]
+
+    return z
+
+
+import numbers
```
gomezzz: at the top please

torchquad/integration/qmc.py
```diff
@@ -91,3 +91,209 @@ def uniform(self, size, dtype):
...
+            if self._seed is not None:
+                gen.manual_seed(self._seed)
+            else:
+                gen.seed()  # a fresh torch.Generator() otherwise keeps a fixed internal
+                # default seed, which would make every unseeded shift
+                # identical instead of independently random
+            shift_values = torch.rand(dim, generator=gen, device=device)
+            points = torch.remainder(points + shift_values.to(dtype), 1)
+
+        return points.to(dtype)
```
gomezzz: `points` is already `dtype` on both branches, I think

torchquad/integration/qmc.py
```diff
@@ -91,3 +91,209 @@ def uniform(self, size, dtype):
...
+
+    z = data[:d, 1]
+
+    return z
+
+
+import numbers
+
+
+def _check_positive_whole_number(value, name):
```
gomezzz: I think the only caller is `MonteCarlo.calculate_sample_points`, which
builds `size=[N, dim]` from values already validated by `BaseIntegrator._check_inputs`.
`Sobol.uniform` handles the same input with `int(size[0]), int(size[1])`. Also, the name says "positive" but the function never checks positivity, and `uniform`'s
`Raises:` documents only `ValueError` while this path raises `TypeError`.

Samuel-Vangu:
Hi @gomezzz ,

Thanks for the thorough review! Addressed all 9 points . All existing tests pass.  . Can you redo a review when you get a chance ? Happy to re-address anything :

1. Converted the generating vector file from plain text (9125 lines) to
   compressed `.npz`; added the missing `package-data` entry in
   pyproject.toml (verified with `python -m build --sdist`). No git-lfs
   needed at this size once it's binary.
2. Split the load into a cached `_load_full_generating_vector` (via
   `functools.lru_cache`) and a thin per-call slice.
3. Added a `backend` constructor arg mirroring `Sobol`, validated in
   `__init__` with a clear `ValueError`, documented, and tested.
4. Added `_MIN_POINTS`, `_MAX_POINTS`, `_LATTICE_FILE` module constants;
   the dimension bound is now derived from the loaded array's shape
   rather than hard-coded.
5. Changed the non-power-of-2 case from `warnings.warn` to `raise
   ValueError`, updated the docstring and the corresponding test.
6. Converted `load_lattice_vector`'s docstring to Google style.
7. Moved `imports to the top; `torch` (and `scipy` in
   `Sobol`) stay local since neither is a hard dependency.
8. Removed the redundant trailing `.to(dtype)`.
9. Removed `_check_positive_whole_number`; `uniform` now mirrors `Sobol`
   with plain `int(size[0])`, `int(size[1])`, since the only real caller
   already guarantees clean ints.

Samuel-Vangu:
Hi @gomezzz,

Thanks for the feedback! Addressed all three points:

1. **Tests** — `Lattice`'s test now goes through `compute_integration_test_errors` against the shared analytical test collection from `helper_functions.py`. Runs per backend (`test_lattice_collection_numpy/torch/tensorflow/jax`).

2. **Backend** — implemented the `anp`-based construction as you suggested; `Lattice` now supports all four backends (numpy, torch, tensorflow, jax), each covered individually by the accuracy, beats-mc, determinism, and collection tests.

3. **Baker's transform** — added the optional `tent` flag as suggested. One difference: I used `anp.abs` instead of `torch.abs` to keep it backend-agnostic, consistent with point 2. Added `test_lattice_tent_improves_nonperiodic_convergence`, which shows tent=True is on average >20x more accurate than tent=False on a non-periodic integrand over 15 seeds.

All 34 tests pass.

Happy to re-address anything — let me know when you get a chance to take another look.

sylvesterkaczmarek (commented):
`tent=True` can violate the sampler's `[0, 1)` range. With the bundled generating vector's first component `z=1`, `N=1024` and `shift=False`, point `i=512` is exactly `0.5`, so the tent map returns `1.0`. Either define/handle the tented range as `[0, 1]` or keep it half-open, and add an unshifted midpoint regression.

Samuel-Vangu:
Hi @sylvesterkaczmarek                                                                                                                                                                                                                                                                                                                                                                                                                    

Confirmed -- and not rare: with z1=1 (the standard convention in the
bundled table) and N always a power of 2, index N/2 always lands on 0.5,
where the tent map's peak is exactly 1.

Went with accepting [0, 1] for tent=True rather than clipping. No code
change needed -- the transform was already correct, just undocumented.
I also proved the non-tented case (shifted or not) is provably confined
to [0, 1): the unshifted formula is bounded by (N-1)/N, and the shift's
final mod 1 always lands in [0, 1) by definition. So [0, 1) stays an
exact guarantee everywhere except tent=True.

Documented this in the docstrings and added the regression test at
N=1024, i=512.

sylvesterkaczmarek (approved):
The tented sampler contract is now documented as [0,1], while the untented path retains the exact [0,1) guarantee, and the N=1024 midpoint regression pins the legitimate tent-map value 1.0. That resolves the contract mismatch I raised.

PR #250, opened 2026-07-30
feat(qmc): add Halton quasi-Monte Carlo sampler

Samuel-Vangu:
# Description

Summary of changes
* Adds `Halton`, a (optionally scrambled) Halton low-discrepancy sampler, shaped exactly like `Sobol` and `RNG` — plugs into `MonteCarlo.integrate` via the existing `rng=` slot, no changes to `MonteCarlo` itself.
* `torch` backend: pure PyTorch implementation (digit extraction, hierarchical digit-shift scrambling, integer Horner reconstruction), generated directly on the target device — no CPU/NumPy round-trip needed, unlike `SobolEngine`.
* `numpy` / `jax` / `tensorflow` backends: `scipy.stats.qmc.Halton` (already a hard dependency), converted to the requested backend.
* Unlike Sobol, does **not** require the sample size to be a power of two.
* Adds `tests/halton_test.py`, mirroring `sobol_test.py`'s structure, plus Halton-specific regression tests (see Test plan).

## Resolved Issues
- [x] Addresses #140 (MonteCarlo path for Halton; VEGAS + JIT support out of scope here, same as for Sobol)

## How Has This Been Tested?
- [x] `halton_test.py`, all backends: analytic accuracy against the test-function collection, Halton < plain-MC error at equal N (fixed seed), seed reproducibility, torch gradient flow
- [x] `test_digits_needed_is_per_dimension`: deterministic, seed/device-independent unit test guarding against a real bug found during development (a shared digit count across dimensions silently broke equidistribution for every dimension except the one with the smallest base)
- [x] `test_halton_column_independent_of_other_dimensions`: exact-equality check that a dimension's column never depends on which other dimensions are requested alongside it
- [x] `test_halton_matches_scipy_when_unscrambled`: pure-PyTorch and SciPy backends agree to within float64 machine epsilon (~2e-16) when scrambling is off, validating the digit extraction/reconstruction independently of the scrambling logic
- [x] `test_halton_no_power_of_two_warning`: confirms no spurious warning for non-power-of-two N
- [x] `ruff format --check .` clean (57 files)

## Related Pull Requests
- #244 (Sobol) — same issue, same sampler shape/interface, same testing pattern

### DIFF
--- a/tests/halton_test.py
+++ b/tests/halton_test.py
@@ -1,0 +1,318 @@
+"""Tests for the Halton quasi-Monte Carlo sampler.
+
+Halton points plugged into ``MonteCarlo`` via the ``rng`` slot must (a)
+integrate the whole analytic test-function collection accurately, (b) beat
+plain pseudo-random Monte Carlo at the same sample count, and (c) reproduce
+bit-for-bit for a fixed seed -- exactly the same contract as Sobol (see
+``test_sobol.py``), whose collection/beats-mc/determinism/gradient tests are
+mirrored below with Halton-appropriate bounds and sample sizes.
+
+On top of that, this file adds tests for properties that are SPECIFIC to
+Halton and worth guarding against regression, because each of them corresponds
+to a real bug found and fixed during development of the torch backend:
+
+    - no power-of-two requirement (unlike Sobol);
+    - the (0,m,1)-net stratification property, verified independently per
+      dimension/base -- this is the exact test that caught an early bug where
+      a single digit count `m` shared across all dimensions (driven by the
+      smallest base) silently broke equidistribution for the other,
+      larger-base dimensions;
+    - agreement between the pure-PyTorch construction and SciPy's reference
+      implementation when scrambling is OFF: since both compute the exact
+      same classical van der Corput sequence, they should match to within
+      float64 machine epsilon, which is a strong end-to-end check that the
+      pure-PyTorch digit extraction and reconstruction is mathematically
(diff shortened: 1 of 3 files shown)


Samuel-Vangu:
@gomezzz, this one's ready for CI too whenever you get a chance — thanks!

Samuel-Vangu:
@gomezzz ,

While waiting on Lattice, I checked Halton against everything you flagged
there and fixed what was necessary.

19/19 tests pass.
Happy to have you take a look whenever you're free.

sylvesterkaczmarek (commented):
The torch scrambling is truncated to `_digits_needed(base, number_of_points)`, so its claimed unbiased randomisation is not actually uniform: for `N=1` in the first dimension, `m=1` and the sample can only be `0` or `0.5` across seeds. That biases the QMC estimator for general integrands; add a random tail/continuous shift (or otherwise make the randomisation unbiased) and include an `N=1` distribution regression.

Samuel-Vangu:
@sylvesterkaczmarek 

Confirmed and fixed. Added a continuous tail shift after digit
scrambling.

Added the N=1 regression: 200 seeds now span way more than 2 distinct
values, and are no longer confined to {0.0, 0.5}.

sylvesterkaczmarek (approved):
The scrambled Halton construction now adds continuous tail randomization, and the N=1 multi-seed regression demonstrates that the first coordinate is no longer confined to the two-point set caused by the truncated digit permutation. This resolves the bias issue I raised.

